PR #15, opened 2020-11-07
Update stdlib versions and 'should' lib over renamed 'except'

mankyKitty:
Messing around with gleam a bit. :)


lpil (approved):
Fantastic! Thank you!

PR #52, opened 2021-06-15
Make examples compile with gleam 0.15.1 and recent stdlib

michallepicki:
When there is a new release of gleam_otp, this should also work with 0.16 I think


lpil (approved):
Wonderful! Thank you

PR #53, opened 2021-06-19
Update deps and fix two-fer example

michallepicki:
Hi!

I updated gleam_otp in the forth exercise to 0.1.5 for compatibility with gleam 0.16. I also changed the stdlib dependency to 0.16 everywhere and loosened the dependency specs with `~>` to let rebar get point releases with fixes when available.

I checked all example implementations and fixed one of them to compile with newer gleam.

### DIFF
--- a/exercises/practice/forth/rebar.config
+++ b/exercises/practice/forth/rebar.config
@@ -8,6 +8,6 @@
 {project_plugins, [rebar_gleam]}.
 
 {deps, [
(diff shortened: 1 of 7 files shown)


michallepicki:
I also fixed the wrong `resistor_color` project name in the gleam.toml file, which was causing this issue to show up: https://github.com/gleam-lang/rebar_gleam/issues/7

lpil (approved):
Wonderful! Thank you!!

PR #75, opened 2021-12-07
Updates hello-world

scripttease:
If this is good I'll update the other exercises and link to issue 72?


lpil (approved):
Perfect! Thank you

PR #76, opened 2021-12-07
Update to Gleam v18 (remove rebar3 tooling)

### LINKED ISSUE #72: Convert exercises to use the new Gleam build tool
This will require the developer to build the main branch version of Gleam if this work is started prior to v0.18 being released.

We will no longer be using eunit for the tests so the test functions will need to explicitly be called, or a testing library will need to be used.

scripttease:
edit 8/12/21: This has a workaround in the `forth/example.gleam' for the compiler bug issue 1379 (in gleam-lang/gleam)
All examples updated to gleam 0.18.0.
closes #72 

...

will close #72 after I check the example solutions for changes from v17.


scripttease:
@lpil how do I request reviews? Or should I just wait for a maintainer to spot PRs? (It is not available as an option for me, I suspect I need a different permission, should I just tag you or just wait on it being spotted? Either is fine! Thanks)

lpil (approved):
Fab! Thank you

PR #90, opened 2022-09-09
Add difference of squares exercise

nono:
cc: @lpil


lpil (approved):
Perfect! Thank you very much

PR #91, opened 2022-09-12
Add accumulate exercise

nono:
cc: @lpil


exercises/practice/accumulate/src/accumulate.gleam
```diff
@@ -0,0 +1,3 @@
+// pub fn accumulate(list: List(a), fun: fn(a) -> b) -> List(b) {
+//   todo("Implement this function")
+// }
```
lpil, line 3: How can this is commentted out when in the last one it was not? I would have assumed that we want to stick to one or the other

lpil (approved):
Perfect!!

PR #92, opened 2022-09-14
Fix forth difficulty

nono:
https://exercism.org/docs/building/tracks/new/add-initial-exercises says it is an hard problem.

### DIFF
--- a/config.json
+++ b/config.json
@@ -106,7 +106,7 @@
         "uuid": "8176ac78-d4fd-475f-ac8b-e205ca7b3a77",
         "practices": [],
         "prerequisites": [],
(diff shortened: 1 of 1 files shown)


lpil:
Huh, is this value still needed? I'm not seeing it in other courses

https://github.com/exercism/elixir/blob/main/exercises/practice/forth/.meta/config.json

lpil (approved):
Oh wait! I was looking in the wrong place. Thanks!

nono:
Yes, the value is in the `config.json` that lists all the exercises: https://github.com/exercism/elixir/blob/main/config.json#L2949

PR #93, opened 2022-09-14
Add Pascals triangle exercise

nono:
cc: @lpil

PR #94, opened 2022-09-15
added bowling exercise

massivefermion:


lpil:
Hey @massivefermion ! Thanks for this. Could you also supply a solution for the exercise?

Here's an example with a solution https://github.com/exercism/gleam/pull/93

exercises/practice/bowling/test/bowling_test.gleam
```diff
@@ -0,0 +1,240 @@
...
+
+  rolls
+  |> roll_and_last_roll_be_error(0)
+}
+
+pub fn trying_to_score_game_before_rolling_bonus_rolls_after_strike_in_last_frame_test() {
+  let rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10]
+
+  rolls
+  |> list.each(fn(s) { roll(s) })
```
lpil, line 186: I think this will need to be changed- it is discarding the result of these calls to `roll` so it will have no effect.

massivefermion: Oh...yeah! I think my day job with nestjs and OOP sometimes messes up my mind. Looks like I was expecting an stateful object to be back there. Forgive me for my transgressions against functional programming! :)

massivefermion:
ok, I think it's correct now, but please let me know if I made any more mistakes. I try to provide the solution as soon as possible.

exercises/practice/bowling/test/bowling_test.gleam
```diff
@@ -0,0 +1,236 @@
...
+pub fn rolling_after_bonus_rolls_after_strike_test() {
+  let rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 3, 2]
+
+  rolls
+  |> roll_and_last_roll_be_error(2)
+}
+
+fn roll_and_check_score(rolls: List(Int), correct_score: Int) {
+  rolls
+  |> list.fold([], fn(state, pins) { list.append(state, roll(state, pins)) })
```
lpil, line 219: I think this might be the scan function

https://hexdocs.pm/gleam_stdlib/gleam/list.html#scan

Also it's good practice to never append to a list, it is a very slow operation. Best to always prefix to a list, and if you need it in the other order reverse it at the end.

massivefermion: I don't think scan is the right function to use. In order for the roll function to score each roll, it needs to have access to the number of knocked down pins in previous rolls and also the scores don't accumulate when rolling. Only when the game is done and you want to calculate the total score, you sum over the frames.

massivefermion:
ok, all the tests are passing, it seems to be done!

lpil (approved):
Thank you!!

PR #95, opened 2022-10-01
Rejig formatting

lpil:
Removing some pipes for legibility and using pattern matching in a few places.

### DIFF
--- a/exercises/practice/bowling/.meta/example.gleam
+++ b/exercises/practice/bowling/.meta/example.gleam
@@ -22,69 +22,41 @@ fn score_rolls(rolls: List(Int)) -> Int {
 }
 
 fn is_frame_rolls_valid(frame: Frame) -> Bool {
-  frame.rolls
-  |> score_rolls <= 10
+  score_rolls(frame.rolls) <= 10
 }
 
 fn is_frame_open(frame: Frame) -> Bool {
-  frame.rolls
-  |> list.length == 2 && frame.rolls
-  |> score_rolls < 10
+  list.length(frame.rolls) == 2 && frame.rolls(score_rolls) < 10
 }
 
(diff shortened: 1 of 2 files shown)


massivefermion (approved):

exercises/practice/bowling/.meta/example.gleam
```diff
@@ -22,187 +22,125 @@ fn score_rolls(rolls: List(Int)) -> Int {
...
         True -> Error(GameComplete)
         False -> {
           let new_game =
             Game(
               game.frames
-              |> list.map(fn(f) {
-                f
-                |> add_roll_to_frame(knocked_pins)
-              }),
+              |> list.map(add_roll_to_frame(knocked_pins)),
```
massivefermion, line 123: it seems what you're doing here is `game.frames |> list.map(add_roll_to_frame(knocked_pins))`, which means instead of a function, you're trying to give the return value of `add_roll_to_frame` to `list.map`, except that you're not giving a frame to `add_roll_to_frame` as the first argument. So I think this should be left as it was.

massivefermion (changes requested):
see comment on line 123 in example.gleam

massivefermion:
Sorry, but I'm not sure what we're waiting for here. Should I do something?

lpil:
Hi @massivefermion ! I'm not sure either. I can't merge for some reason, I'm not sure who I am waiting on for a review.

@exercism/gleam anyone else a member of this group? :)

SleeplessByte (approved):

SleeplessByte:
@lpil I'm not but as you're just there by yourself, I do have rights! 

Merged this 🚀 . Feel free to tag me whenever, and / or Erik or Jeremy.

PR #97, opened 2022-10-11
Fix configlet status badge

SleeplessByte:
The configlet workflow name has changed, thus has its reference URL.  Added linkback to the workflow so you can see what went wrong when it has failed.


lpil (approved):
Thank you

Issue #96, opened 2022-10-06
Run example solutions on CI


lpil:
For each exercise
1. Get the solution path from the config.json file
2. Copy the example.gleam file to the solution location
3. Run `gleam test` in that project

It would be good if this script can be run locally.

The script should reset the solution file back to the original contents after the test.

ErikSchierboom:
For inspiration, you could look at https://github.com/exercism/cobol/blob/main/bin/test or https://github.com/exercism/unison/blob/main/bin/test

lpil:
Cool! Thank you

PR #99, opened 2022-11-06
Run example solutions on CI

lpil:
Closes https://github.com/exercism/gleam/issues/96

Closes https://github.com/exercism/gleam/issues/28

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -20,6 +20,8 @@ set -euo pipefail
 # Example:
 # ./bin/test
 
+exit_code=0
+
 function verify_exercise {
   local dir slug implementation_file stub_backup_file module
   dir=$(realpath "$1")
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Synopsis:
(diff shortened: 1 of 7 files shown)


bin/test
```diff
@@ -0,0 +1,66 @@
...
+# ./bin/test
+
+exit_code=0
+
+function verify_exercise {
+  local dir slug implementation_file stub_backup_file module
+  dir=$(realpath "$1")
+  slug=$(basename "${dir}")
+  module=${slug//-/_}
+  implementation_file="${dir}/.meta/example.gleam"
```
SleeplessByte, line 30: In the future, you may want to pull this from the config file, because it's commonplace to use `example` (or `proof.ci`) for practice exercises and `exemplar` for concept exercises.

SleeplessByte (approved):

bin/test
lpil: I thought that but the other scripts I referenced worked this way, so I figured it would be easier to do this rather than remember how jq works 😁

SleeplessByte: Yes, most tracks hardcoded it too 😄, but some like JavaScript started using the config when provided for more flexibility.

PR #100, opened 2022-11-06
Check Gleam code format

lpil:


.github/workflows/test.yml
```diff
@@ -21,3 +21,6 @@ jobs:
 
       - name: Run tests for all exercises
         run: bin/test
+
+      - name: Check formatting
+        run: gleam format --check exercises/*/*/{src,test,.meta}/*.gleam
```
SleeplessByte, line 26: Does this mean that you won't expect folders inside the exercise directories? Definitely not a problem, but just want to make sure that's communicated somewhere or at least in your head for when you eventually do 😄 .

SleeplessByte (approved):

.github/workflows/test.yml
lpil: I'll add support for that too

SleeplessByte: Out of the 200+ exercises on the JS and TS track it's like 5 that have folders. I also had the exact same setup at first and those files remained unformatted 🥹😅

PR #102, opened 2022-11-06
Fix glob

lpil:
I'm a fool and merged too early. I don't know why globbing like this does not work here, I guess GitHub actions uses a different shell syntax to my machine.


SleeplessByte (approved):
It should work, but perhaps it does force at least one directory.

🤔

PR #103, opened 2022-11-06
added anagram exercise

massivefermion:


lpil (commented):
Great stuff! Thank you.

I think we could simplify this and also make it run faster by not generating all possible anagrams of the word to check against, instead having something like this:

```gleam
candidates
|> list.filter(is_anagram(_, word))
```

where `is_anagram` could check that the word and the candidate if their letters are sorted into alphabetical order.

massivefermion:
good idea, I changed it.

lpil (approved):
Really nice! Thanks!

PR #106, opened 2022-11-10
added roman numerals exercise

massivefermion:
I already had the solution in [blah](https://hex.pm/packages/blah), so...


lpil (approved):
Wow that's a really cool way to solve this, I wouldn't have thought of that. Thank you!

PR #110, opened 2022-11-16
Ensure manifests are correct

lpil:
Hopefully this fixes https://github.com/exercism/gleam-test-runner/issues/3 !


lpil:
I don't have an amd64 machine to test with I'm afraid.

lpil:
That would be a considerable amount of work. I'd need to create aarch64 images for Erlang, Gleam, and exercism/gleam-test-runner. I'm afraid I don't have time for that at present and also GitHub actions doesn't support aarch64 so there's no way to build and test these images.

ErikSchierboom:
@lpil One thing to note is that students won't see any exercise update for changes to either `toml` file. This is because these files are not included in our "is an exercise out of date" check. To fix this, you'll need to add these filenames to the `files.invalidator` key in the `.meta/config.json` file of each exercise. 

What I recommend you do is to add them to the track config.json first, and then use `configlet sync --filepaths --update --yes`.

lpil:
What are the rules around whether an exercise is out of date? If there are no breaking changes with the dependencies is the track still out of date?

ErikSchierboom:
An exercise is out of date when one of its "significant files" (files.solution, files.test, files.editor, files.invalidator) has changed since the student last updated (or started) the exercise.

lpil:
Sorry, I mean more how do we decide if something should go into `files.invalidator`? I am expecting the `gleam.toml` and `manifest.toml` files to change every month or two but it not to be a breaking change to the exercises at all.

Would be confusing for students for their submissions to be marked as out of date when nothing has changed from their point of view?

ErikSchierboom:
> Sorry, I mean more how do we decide if something should go into files.invalidator? I am expecting the gleam.toml and manifest.toml files to change every month or so but it not to be a breaking change to the exercises at all.

It isn't shown as a breaking change in the interface, but merely as an upgrade.

> Would be confusing for students for their submissions to be marked as out of date when nothing has changed from their point of view?

I think what would be more confusing is for students to see that their exercise might not work on the latest test runner due to outdated dependencies. We also have an option (enabled by default) that automatically updates exercises to the latest version, so students wouldn't even see the update message.

lpil:
Wonderful! Thank you

PR #116, opened 2022-11-18
added dominoes exercise

massivefermion:


config.json
```diff
@@ -222,14 +222,22 @@
...
           "math"
         ]
+      },
+      {
+        "slug": "dominoes",
+        "name": "Dominoes",
+        "uuid": "b59b630e-8648-4498-b8d8-fcfaf261ece3",
+        "difficulty": 7,
+        "practices": [],
+        "prerequisites": []
```
lpil, line 240: Could you fill these in please? Thanks!

lpil (commented):
Fab, thank you!

There's a lot of reversing and accessing the ends of list here. These are very slow operations that should best be avoided. I think the queue data structure would potentially be a better fit here as it can be quickly accessed from both ends.

Can we also remove the use of `assert`? Best to avoid that as much as possible.

massivefermion:
I changed things as you requested and everything seems to be fine. But for some unknown reason the checks are failing!

lpil (commented):
Yup that's right, the test runner is not on the latest version of Gleam yet. Probably easiest to not use `use` just yet for this.

massivefermion:
ok, done!

lpil (approved):
You star, thank you!

PR #125, opened 2023-01-19
Protein translation

lpil:


chouzar (approved):
This was a really fun exercise to work through gleam, I stumbled a bit on the program expectations. 

While this is implicit in the result of the `proteins` function, maybe we should make emphasis into:
* We need a list of named proteins `List(String)`.
* There can be correct and incorrect sequences `Result(List(String))`.
* Not all sequences need to have a stop codon to be succesful. 

An maybe some brief examples:

```gleam
proteins("AUGUUU")
// Ok(["Methionine", "Phenylalanine"])

proteins("ABC")
// Error(Nil)

proteins("AUGUUUUGA")
// Ok(["Methionine", "Phenylalanine"])
```

Is there a place to explain gleam concepts? For this exercise I'm thinking of the `<>` operator, `Result` type and maybe `bitstring` matching.

lpil:
The exercises have standardised documentation as far as I understand, I'm not sure if we're supposed to add to it or not.

> Is there a place to explain gleam concepts? For this exercise I'm thinking of the <> operator, Result type and maybe bitstring matching.

Aye in future we'll have `concepts` exercises as well as `practice` ones. None of those have been made yet, I'm hoping to acquire funding to produce them.

jiegillet:
I can confirm concept exercises are the best way to add explanations, practice exercises are best kept lean.

chouzar:
Got it thanks for clarifying!

PR #129, opened 2023-01-19
Add Test documentation

lpil:
Part of https://github.com/exercism/gleam/issues/63

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -11,8 +11,7 @@ This will execute the test file found in the `test` subfolder -- a file ending i
 ## Test functions
 
 The tests use a Gleam test framework named [Gleeunit](https://github.com/lpil/gleeunit).
-Any public function in the `test` directory with a name ending in `_test` will be
-considered a test by Gleeunit.
+Any public function in the `test` directory with a name ending in `_test` will be considered a test by Gleeunit.
 
 For example:
 
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,2 +1,27 @@
-# Tests
+# Running tests
 
+From the terminal, change to the base directory of the exercise then execute the tests with:
+
+```bash
+$ gleam test
+```
+
+This will execute the test file found in the `test` subfolder -- a file ending in `_test.gleam`
+
+## Test functions
+
+The tests use a Gleam test framework named [Gleeunit](https://github.com/lpil/gleeunit).
+Any public function in the `test` directory with a name ending in `_test` will be
+considered a test by Gleeunit.
+
+For example:
+
+```gleam
+import bob
+import gleeunit/should
+
+pub fn shouting_test {
+  bob.hey("WATCH OUT!")
+  |> should.equal("Whoa, chill out!")
+}
+```


docs/TESTS.md
```diff
@@ -1,2 +1,27 @@
...
+$ gleam test
+```
+
+This will execute the test file found in the `test` subfolder -- a file ending in `_test.gleam`
+
+## Test functions
+
+The tests use a Gleam test framework named [Gleeunit](https://github.com/lpil/gleeunit).
+Any public function in the `test` directory with a name ending in `_test` will be
+considered a test by Gleeunit.
```
jiegillet, line 15: In general, Exercism tries to stick to one line per sentence in markdown docs.
```suggestion
Any public function in the `test` directory with a name ending in `_test` will be considered a test by Gleeunit.
```

jiegillet (approved):
Nice.

Just asking, have you setup a mechanism in the test runner to run only one test at a time and skip the others? It's Exercism's approach to TDD.
If you did, it might be worth mentioning here.

jiegillet:
Ok, it looks like I don't have approving rights.
@ErikSchierboom could I ask you to add me to the list of Gleam maintainers? I would like to help out for a while.

docs/TESTS.md
```diff
@@ -1,2 +1,27 @@
-# Tests
+# Running tests
 
+From the terminal, change to the base directory of the exercise then execute the tests with:
```
chouzar, line 3: Wondering if it would be useful to show navigation to the `bob` directory here:

```bash
cd exercises/practice/bob
```

However I'm not entirely sure if end users will have the same directory structure as above.

chouzar (approved):

docs/TESTS.md
lpil: OK! I copied from some other track which did not do this.

docs/TESTS.md
lpil: The user wouldn't have that structure when they have downloaded the exercise locally so likely would be confusing.

lpil:
@jiegillet We've not- it's not a feature that the Erlang or Gleam test runners have.

docs/TESTS.md
jiegillet: No, it's not the same directory structure. 
When they install the exercise with the exercism CLI they will see
```bash
➜  ~ exercism download --exercise=hello-world --track=typescript

Downloaded to
/Users/jie/Documents/exercism/jie-exercism-tracks/typescript/hello-world
```

jiegillet:
> @jiegillet We've not- it's not a feature that the Erlang or Gleam test runners have.

I see. Maybe it could be done artificially by renaming the tests `_test_skip` or something. But that's a big decision, not for this PR.

lpil:
I think that would not work as it would immediately say that the user has passed the exercise once they solve the first test. Without the pending indicator they wouldn't know there's more work still to do.

ErikSchierboom:
> Ok, it looks like I don't have approving rights.
@ErikSchierboom could I ask you to add me to the list of Gleam maintainers? I would like to help out for a while.

I can't, but @ihid can.

iHiD:
@jiegillet Done!
@lpil, Jie is very experienced Exercism maintainer so can be someone you can bounce things off! 🙂

jiegillet:
Thank you!

ErikSchierboom:
Thanks for the help @jiegillet!

PR #133, opened 2023-01-20
Sync all metadata, docs and tests

jiegillet:
I ran the following 3 commands
```bash
➜  gleam git:(jie-sync) ✗ bin/configlet sync --metadata --update --yes    
➜  gleam git:(jie-sync) ✗ bin/configlet sync --docs --update --yes    
➜  gleam git:(jie-sync) ✗ bin/configlet sync --tests --update
```

For the metadata and docs, it's best to trust configlet.

For the tests, the command goes through all exercise, checks `tests.toml` and asks if each exercise missing has been implemented. I checked the test files and answered as accurately as I could (there were a lot of tests).
In general, this data is very useful for when test data gets updated upstream, it makes the track updates very simple.

When tests are not implemented, they are marked as `include = false` in the file. Sometimes there are good reasons to not implement a case (for example, some test cases check for immutability, which makes no sense in Gleam) and in those cases, it's good practice to leave a reason via `comment = "Gleam is immutable"`.

I noticed a bunch of cases were missing (I'm guessing some test files were copied from some other tracks with missing data?) but I didn't implement them or add justifications. We might want to open an issue to track that if you want to address that at some point (it's low priority though).

### DIFF
--- a/exercises/practice/accumulate/.docs/instructions.md
+++ b/exercises/practice/accumulate/.docs/instructions.md
@@ -1,9 +1,6 @@
 # Instructions
 
-Implement the `accumulate` operation, which, given a collection and an
-operation to perform on each element of the collection, returns a new
-collection containing the result of applying that operation to each element of
-the input collection.
+Implement the `accumulate` operation, which, given a collection and an operation to perform on each element of the collection, returns a new collection containing the result of applying that operation to each element of the input collection.
 
 Given the collection of numbers:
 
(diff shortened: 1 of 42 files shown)


lpil (approved):
> I noticed a bunch of cases were missing (I'm guessing some test files were copied from some other tracks with missing data?) but I didn't implement them or add justifications. We might want to open an issue to track that if you want to address that at some point (it's low priority though).

I think that is the case. I agree, it would be good to have these, issues would be grand.

PR #142, opened 2023-01-22
Update two-fer to match canonical data

kytrinyx:
This updates two-fer to match the canonical specs in the problem-specifications repository.

I hesitated about making it Alice instead of Gilberto Barros
https://github.com/exercism/problem-specifications/blob/51fcbc35abc4ac8d0394d3186f10b2fed04ef3ce/exercises/two-fer/canonical-data.json#L18
because there's no reason why it can't be Gilberto, and I like the creativity of it.

I'd be happy to put that back if you prefer it.


jiegillet (approved):
Thank you!
I think we can live with that change, it improves maintainability!

Issue #134, opened 2023-01-20
Add practice exercise missing test cases


jiegillet:
When running `bin/configlet sync --tests --update`, configlet goes through each practice exercise, checks if each test case has been implemented and logs the results in `exercises/practice/<slug>/.meta/tests.toml`.
In general, this data is very useful for keeping test cases in sync with the `[problem-specifications](https://github.com/exercism/problem-specifications)` repository .

When tests are not implemented, they are marked as `include = false` in the file. Sometimes there are good reasons to not implement a case (for example, some test cases check for immutability, which makes no sense in Gleam) and in those cases, it's good practice to leave a reason via `comment = "Gleam is immutable"`.

There are currently a bunch of cases missing without justifications.
We should go through each exercises with missing cases and either implement them or add a justification.
Of course, it can be done in chunks.

PR #144, opened 2023-01-22
Update anagram to match canonical data

kytrinyx:
This updates the anagram tests to match the tests listed in the
tests.toml. This is taken from the canonical-data.json file in
the problem-specifications repo.


jiegillet (approved):
Thank you.
These test cases made my head spin when I checked which were implemented and which were not. We have an issue open #134 about missing test cases, I'm hoping we can fill in the missing ones at some point.

kytrinyx:
Ah, nice! I haven't started looking at open issues yet. I totally forgot to look for "include=false" in the tests.toml. Let me double-check the exercises that I've already gone through!

PR #145, opened 2023-01-22
Remove `darts` solution from stub

jiegillet:
This exercise is easy enough, let's not make it _too_ easy :p


kytrinyx (approved):
😀 Ah, yes. This is a good shout.

lpil:
Oops 😅

PR #146, opened 2023-01-22
Reset tests.toml for anagram exercise

kytrinyx:
I updated the anagram test suite in #144 to match the
canonical-data.json file more closely, but I wasn't aware of
the open issue #134, so I didn't put in the tests that said
they were ignored.

This removes the 'include = false' from the tests.toml (except
in the case of re-implemented tests) and adds the missing test
cases.


jiegillet (approved):
Beautiful, thank you!

PR #147, opened 2023-01-22
Add practice exercise generator

jiegillet:
Having a generator makes it way, way easier to add exercises, so I thought I would port the one I wrote for the Elm track into Gleam.

Run `bin/generate_practice_exercise.sh <exercise-slug>` and it will create all of the files required for the exercise (minus `tests.toml`, which configlet can do). It works by parsing the JSON of the canonical data, and creating all test cases and solution stub file.

It's not perfect by any means, the biggest advantage is that all the test cases will be in the test file. The data usually requires a lot of regex search-and-replace to get it in a proper shape though. It also tries to guess the types, but that doesn't work that well for non-trivial cases.

I tested the script on a dozen of random exercises, but it's still possible that one exercise could break the script (like a strange variable name in the JSON that is illegal in Gleam or something).

Once concept exercises are being built, it won't be hard to adapt the script to generate all the required files (but it doesn't help nearly as much since there are no test cases).

I'm still a Gleam novice, so the code is probably not perfect. In particular, the JSON parsing game me trouble, I expected to find all the decoders I needed in `gleam_json`, but it only has encoders, so I used `gleam/dynamic` but it doesn't have things like Elm's `succeed`, `lazy` or `andThen`.


bin/generate_practice_exercise.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
lpil, line 1: There's a mix of 2 and 4 space indentation in this file. Is that an accident?

exercise_generator/gleam.toml
```diff
@@ -0,0 +1,11 @@
...
+version = "0.1.0"
+description = "Test and stub file generator for the Exercism Gleam track"
+
+[dependencies]
+gleam_stdlib = "~> 0.26"
+gleam_erlang = "~> 0.17"
+gleam_json = "~> 0.5"
+
+[dev-dependencies]
+gleeunit = "~> 0.9"
```
lpil, line 11: Can probably just drop the test code as it is unused.

exercise_generator/src/exercise_generator.gleam
```diff
@@ -0,0 +1,447 @@
...
+
+fn json_data_to_gleam_type(data: JsonData) -> String {
+  case data {
+    JsonNull -> "Result(unknown, Nil)"
+    JsonBool(_) -> "Bool"
+    JsonInt(_) -> "Int"
+    JsonFloat(_) -> "Float"
+    JsonString(_) -> "String"
+    JsonList([]) -> "List(unknown)"
+    JsonList([head, ..]) -> "List (" <> json_data_to_gleam_type(head) <> ")"
```
lpil, line 66:
```suggestion
    JsonList([head, ..]) -> "List(" <> json_data_to_gleam_type(head) <> ")"
```

lpil (approved):
Very nice! This would have saved me a lot of time 😅

bin/generate_practice_exercise.sh
jiegillet: Indeed, copy/paste mixup.

PR #149, opened 2023-01-22
Update roman-numerals tests

kytrinyx:
This removes the include=false from tests.toml, as it seems like
the exclusions were probably accidental.

It then adds the missing tests, updating the test names to match
the descriptions in the canonical data.

### DIFF
--- a/exercises/practice/roman-numerals/.meta/config.json
+++ b/exercises/practice/roman-numerals/.meta/config.json
@@ -3,7 +3,8 @@
     "massivefermion"
   ],
   "contributors": [
-    "lpil"
+    "lpil",
+    "kytrinyx"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/roman-numerals/.meta/tests.toml
+++ b/exercises/practice/roman-numerals/.meta/tests.toml
@@ -71,7 +71,6 @@ description = "16 is XVI"
 
 [4465ffd5-34dc-44f3-ada5-56f5007b6dad]
 description = "66 is LXVI"
-include = false
 
 [902ad132-0b4d-40e3-8597-ba5ed611dd8d]
 description = "166 is CLXVI"
@@ -84,8 +83,6 @@ description = "1666 is MDCLXVI"
 
 [3bc4b41c-c2e6-49d9-9142-420691504336]
 description = "3001 is MMMI"
-include = false
 
 [4e18e96b-5fbb-43df-a91b-9cb511fe0856]
 description = "3999 is MMMCMXCIX"
-include = false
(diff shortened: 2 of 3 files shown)


jiegillet (commented):
Thank you so much!
I have two comments.

You removed some test cases because they were not in the canonical data (like 32 or 39).  In general I prefer sticking to the test cases, but in this case, it's really easy to verify that the tests are correct, so I think it doesn't hurt to leave them. What do you think?

Second, I think one nice feature of this test files is that the numbers were ordered numerically, I find it nicer for TDD (even though we don't have a skip mechanism). Could we keep the order? I think it's mostly ordered, but 16 is way down there.

kytrinyx:
It's easy to leave them in, and I agree that it doesn't hurt, since once they pass they pass. I'll add them back.

For the ordering I agree. I follow the ordering of the canonical-data. I'm going to go ahead and submit a PR upstream to problem-specifications for this, as it's a nice touch and doesn't actually affect any tracks that already have implemented it.

kytrinyx:
I've opened the PR upstream: https://github.com/exercism/problem-specifications/pull/2185

I just took a look at the extra cases, and I'm not actually sure how I would include them programmatically. Usually for track-specific tests my generator just sticks them at the bottom of the test suite. But here we want them to be ordered numerically. I need to think about how I want to implement that.

jiegillet:
Nice idea with the reordering PR 👍 

For the reordering, if your generator sticks them at the end, may I suggest a good old-fashioned copy-and-paste insert sort? Speaking for myself, it would take me 1/10th of the time I would spend coding it :D

kytrinyx:
Yah, most of my generator is really dumb, and I have (potential) manual tweaks all over the place to avoid spending a bajillion years coding things that are not worth it. For this one I might end up going down that route :)

kytrinyx:
I merged the problem-specifications PR that reorders the test cases, regenerated, and have pushed it up.

Having slept on the bit about extra test cases (for 32, 39, 128, 789, 1776, 1918, and 1954) my preference would be to leave them out, unless they are covering cases that really should be added upstream. While including them doesn't hurt from the student's perspective, it also doesn't really help. Importantly, it reduces maintainability, as the test suite cannot be updated just by rerunning the generator.

Thoughts?

jiegillet:
In my experience syncing exercises in the Elm and Elixir track, I usually find it easier to include the new test cases that configlet tell me about by hand. My generators on my tracks are another type of dumb and discard everything in the file, so except for the simplest of exercises that are generated correctly the first time, it's not worth using, because many have helper functions, etc.

However for a new track, it might be good to stick 100% with upstream, it really does make maintaining easier, so I'm fine with removing the extra test cases.

Is your generator available somewhere? I'd love to take a look.

kytrinyx:
Cool, thank you, @jiegillet. I've emailed you so we can chat about generators outside this thread :)

I've gone ahead and made this ready for review, leaving off the extra cases.

jiegillet (approved):
Looks great, thank you 🏅

PR #151, opened 2023-01-22
Add practice exercise `isogram`

jiegillet:
I thought I would test out the generator with an easy one (and I actually found a bug haha).


kytrinyx:
This adds a lot of files. My expectation would have been to see only new files in `exercises/practice/isogram/*` and one change to the top-level `config.json`. Would you double-check to make sure this is the way you want it? (I don't know gleam so I'm probably missing context).

lpil:
I think the .gitignore is missing here!

jiegillet:
Oh no, you are right, I messed that up.
The .gitignore of the generator is supposed to ignore all those build files, but it's not committed on this branch yet, and I used `git add .` by mistake.

and yes, I forgot the `.ignore` file in the generation!

jiegillet:
Oh no, you are right, I messed that up.
The .gitignore of the generator is supposed to ignore all those build files, but it's not committed on this branch yet, and I used `git add .` by mistake.

and yes, I forgot the `.ignore` file in the generation!

exercises/practice/isogram/test/isogram_test.gleam
```diff
@@ -0,0 +1,77 @@
+import gleeunit
+import gleeunit/should
+import isogram
```
kytrinyx, line 3: I noticed that for existing exercises where the solution only has one function, the import pulls in just the function, like so:

```
import anagram.{find_anagrams}
``` 

Is this a Gleam best practice that we should follow here?

jiegillet: I'm not sure, being a Gleam novice myself.
In general though, I much prefer qualified functions rather than imported, it really helps with readability. Except for types, I do import types.

lpil: I think the general preference is to keep them qualified rather than importing functions directly, but it's generally up to the programmer to decide which they think is more appropriate for the code they are writing. I'd be happy with either in the exercises as it is quite clear where the functions come from..

lpil (approved):
Very nice! I would be good to use the other assertion style as it produces a better error and in future the compiler will upgrade it to the upcoming test assertion syntax.

PR #152, opened 2023-01-22
Run configlet fmt to normalize configs

kytrinyx:
This normalizes exercise configs, which should help
reduce noise in other PRs later.

### DIFF
--- a/config.json
+++ b/config.json
@@ -273,14 +273,23 @@
           "result",
           "case"
         ],
-        "prerequisites": ["list", "string", "result", "case"],
+        "prerequisites": [
+          "list",
+          "string",
+          "result",
+          "case"
+        ],
         "difficulty": 1
       },
       {
         "slug": "diamond",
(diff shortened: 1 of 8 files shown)


kytrinyx:
@jiegillet I noticed that #151 was changing some exercise configs that were unrelated to the exercise being added.

Unfortunately configlet fmt doesn't format the top-level config, so I grabbed the reformatted file from your isogram PR (#151), and removed the bit about isogram.

lpil (approved):
Thank you. Is a particular tool used for this?

jiegillet:
For the exercise config the tool is `bin/configlet fmt --update --yes`.

For the top config, the generator script I use rewrites the config via `jq`, which also formats the file. Actually there is one annoying side effect/bug, the `"average_run_time": 3.0` always gets transformed to to `"average_run_time": 3` but the Exercism tool using this value requires an float. Haven't found a nice fix for that yet.

PR #153, opened 2023-01-23
Tweak pangram tests per canonical-data

kytrinyx:
This removes an include=false from the tests.toml, as the
exclusion seems to have been unintentional (the new test
passes with no changes to the example solution).

It also switches to use the should.be_false / should.be_true style
that other tests use, rather than assert False / assert True.


exercises/practice/pangram/test/pangram_test.gleam
```diff
@@ -1,61 +1,73 @@
...
 
 pub fn case_insensitive_test() {
   let sentence = "the quick brown fox jumps over with lazy FX"
-  assert False = is_pangram(sentence)
+  is_pangram(sentence)
+  |> should.be_false
 }
 
-pub fn a_m_and_upper_a_m_are_26_different_characters_but_not_a_pangram_test() {
+pub fn a_m_and_a_m_are_26_different_characters_but_not_a_pangram_test() {
```
jiegillet, line 69: I think the previous title was better, lost in generation I suppose

jiegillet (commented):
Nice! Only one thing I would revert.

lpil (commented):
Nice! Could you revert the `should.be_true` be changes please? It produces a worse error that does not include the location of the assertion.

exercises/practice/pangram/test/pangram_test.gleam
kytrinyx: Hm, yeah you're right, they get lost in the generation. I'm going to see if I can improve the test title upstream and regenerate.

PR #154, opened 2023-01-23
Swap assertion style in leap

kytrinyx:
This switches leap to use the assert True style of
assertions rather than the should.be_true style.


jiegillet (approved):
Fantastic!

Issue #148, opened 2023-01-22
assert True vs should.be_true in tests


kytrinyx:
I noticed that the style of assertions differs between tests, and I was wondering if that is done intentionally, or if it's more of a _it doesn't actually matter_ type of thing.

E.g.

```
# exercise: leap
pub fn year_divisible_by_100_not_divisible_by_400_in_common_year_test() {
  leap.is_leap_year(2100)
  |> should.be_false
}
```

```
# exercise: pangram
pub fn case_insensitive_test() {
  let sentence = "the quick brown fox jumps over with lazy FX"
  assert False = is_pangram(sentence)
}
```

If it doesn't really matter, is it okay if I just use one style when working on the exercises? I have a (very stupid) generator that I configure, because working with the `canonical-data.json` file manually is a bit of a headache. I can do different things for each exercise if necessary, but if it doesn't matter it's easier to be consistent.

jiegillet:
I personally like the `should.be_false` style, I think I will include that in my generator too.

kytrinyx:
Cool, I will stick with that, then!

lpil:
should.be_true produces a worse error and will be removed in future so using the assert keyword is preferred. Later it will be possible to write "assert someexpression", at which point we will migrate to that.

jiegillet:
Oops, sorry @kytrinyx

lpil:
Thanks gang 💜

kytrinyx:
What should this look like with the other style assertion?

```
pub fn another_four_tile_test() {
  {
    [#(6, 4), #(6, 2), #(2, 0), #(4, 0)]
    |> arrange
    |> list.length > 0
  }
  |> should.be_true
}
```

jiegillet:
I see some options. In ascending order of personal preference:
```gleam
pub fn one_tile_test() {
  assert True =
    [#(2, 2)]
    |> arrange
    |> list.length > 0
}
```

```gleam
pub fn one_tile_test() {
  assert False =
    [#(2, 2)]
    |> arrange
    |> list.is_empty
}
```

```gleam
pub fn one_tile_test() {
  assert [_, ..] = arrange([#(2, 2)])
}
```
but if I was going with the last one, I would change the `should.equal(0)` tests to
```gleam
pub fn another_impossible_three_tile_test() {
  assert [] = arrange([#(5, 3), #(4, 5), #(2, 3)])
}
```

suraj-mandal:
[...]

PR #157, opened 2023-01-25
Tweak dominoes to match canonical data

kytrinyx:
This deletes some include=false in the tests.toml
per #134.

It also rewords the assertions with the goal of avoiding
the should.be_true style of assertion, per #148.

Finally it normalizes the test names based on the test
descriptions in the canonical-data.json file from
problem-specifications.

### DIFF
--- a/exercises/practice/dominoes/test/dominoes_test.gleam
+++ b/exercises/practice/dominoes/test/dominoes_test.gleam
@@ -1,5 +1,4 @@
 import dominoes.{can_chain}
-import gleam/list
 import gleeunit
 
 pub fn main() {
--- a/exercises/practice/dominoes/test/dominoes_test.gleam
+++ b/exercises/practice/dominoes/test/dominoes_test.gleam
@@ -7,132 +7,69 @@ pub fn main() {
   gleeunit.main()
 }
 
-pub fn empty_test() {
+pub fn empty_input_empty_output_test() {
   []
   |> arrange
   |> list.length
   |> should.equal(0)
 }
 
-pub fn impossible_one_tile_test() {
-  [#(2, 4)]
-  |> arrange
-  |> list.length
-  |> should.equal(0)
+pub fn singleton_input_singleton_output_test() {
+  assert [_, ..] = arrange([#(1, 1)])
 }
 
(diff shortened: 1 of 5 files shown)


exercises/practice/dominoes/test/dominoes_test.gleam
```diff
@@ -7,132 +7,69 @@ pub fn main() {
   gleeunit.main()
 }
 
-pub fn empty_test() {
+pub fn empty_input_empty_output_test() {
   []
   |> arrange
   |> list.length
   |> should.equal(0)
 }
```
jiegillet, line 15: This one could also be in the same style as the others
```suggestion
pub fn empty_input_empty_output_test() {
  assert [] = arrange([])
}
```

jiegillet (commented):
Thank you, this exercise test cases were all over the place...

I looked at the canonical data, and the tested property should actually be `can_chain` and return a Boolean...
I kind of think we should match that, the tests are all binary anyway.

Would you like me to tweak the solution in this PR to implement `can_chain` instead?

exercises/practice/dominoes/test/dominoes_test.gleam
kytrinyx: Oops! My bad :) I had a custom override for that test!

I'm going to leave it for now since we're going to do `can_chain` instead.

kytrinyx:
> I looked at the canonical data, and the tested property should actually be can_chain and return a Boolean...
> I kind of think we should match that, the tests are all binary anyway.

I think that is an excellent idea. I can very easily tweak the test suite, but I've not yet tackled actually writing any code.

I'll push up a (failing) test suite with can_chain in a moment.

kytrinyx:
@jiegillet I made a guess at what the test suite should look at. If you have a better suggestion feel free to give me an example snippet and I can regenerate it.

I didn't touch the example solution. If you're happy to take over from here, that would be lovely :-)

jiegillet:
OK, I've tweaked the solution, the CI passes and the test suite looks great! Good to go.

jiegillet (approved):

kytrinyx:
Yay! Shipping it :)

PR #158, opened 2023-01-25
Add practice exercise `zipper`

jiegillet:
New exercise. I like data structure ones :)


lpil (approved):
This one looks really fun!

PR #159, opened 2023-01-26
Add practice exercise `phone-number`

jiegillet:
This one is great for practicing `Result` chains.


exercises/practice/phone-number/test/phone_number_test.gleam
```diff
@@ -0,0 +1,97 @@
...
+import gleeunit/should
+import phone_number
+
+pub fn main() {
+  gleeunit.main()
+}
+
+pub fn cleans_the_number_test() {
+  phone_number.clean("(223) 456-7890")
+  |> should.equal(Ok("2234567890"))
```
jiegillet, line 11: I tried the `assert Ok("..") = ` versus the `|> should.equal(Ok(".."))` patterns, and I found that the error messages were not equally helpful. 
The `assert` pattern gave me
```
  1) phone_number_test.cleans_numbers_with_dots_test
     Assertion pattern match failed
        value: Error("punctuations not permitted")
     location: phone_number_test.cleans_numbers_with_dots_test:14
     stacktrace:
       phone_number_test.cleans_numbers_with_dots_test
     output: 
```
while the `should` pattern gave me
```
  1) phone_number_test.cleans_numbers_with_dots_test
     Values were not equal
     expected: Ok("2234567890")
          got: Error("punctuations not permitted")
     output: 
```

While the first one game me a location, the second one gave me the expectation directly, which in this case was more helpful, so I went with that.
Once the test runner implements parsing the snippets of code, the situation might be different

lpil: I think in this case it is good either way, though `assert` will continue to improve and we'll remove `should` in future.

exercises/practice/phone-number/.meta/example.gleam
```diff
@@ -0,0 +1,73 @@
+import gleam/string
+import gleam/result
+import gleam/list
+
+pub fn clean(input: String) -> Result(String, String) {
+  Ok(input)
+  |> result.map(remove_valid_punctuation)
+  |> result.then(validate_characters)
```
lpil, line 8: The first bit of wrapping can be skipped
```suggestion
  input
  |> remove_valid_punctuation
  |> validate_characters
```

lpil (approved):
Nice!

exercises/practice/phone-number/test/phone_number_test.gleam
jiegillet: Good to know!

PR #162, opened 2023-01-27
Add practice exercise `binary-search-tree`

jiegillet:
Another nice little one. The generator really makes it easy :)


lpil (approved):
Very slick! Thanks!

PR #163, opened 2023-01-28
forth test cleanup

bcpeinhardt:
Some of the tests weren't running because they weren't named properly (the function names have to end in _test).
I've also rewritten the test "defining_words_with_odd_characters_test" because it was the only test which
required two "statements" on the same line, but I don't think that's what it meant to test.


lpil (approved):
Thank you!

This is such an easy mistake to make, good evidence that we need to revisit how tests work in Gleam

bcpeinhardt:
Idk, I kinda like the whole `_test` convention. In rust I'm constantly forgetting to write `#[test]` or `#[tokio_test]` and I can't ever remember the compiler or clippy asking me if I meant to make the method a test. I don't think it's a mistake I wouldn't make in another language haha.

PR #164, opened 2023-01-30
Add practice exercise `pythagorean-triplet`

jiegillet:
This one might require some thought.

This is an exercise that, solved with a naive approach, might run for longer than the Exercism time limit of 10 seconds because of the last test.

There are a few problems like that, I can think of a couple of approaches:
- some languages can tag a test as slow, and the test runner skips it. I don't think Gleam can do this yet.
- we can leave some hints in `introduction.append.md` on how to find an efficient solution
- we can comment the last test and recommend that students try it locally in `introduction.append.md`


exercises/practice/pythagorean-triplet/.meta/example.gleam
```diff
@@ -0,0 +1,64 @@
...
+  Triplet(factor * a, factor * b, factor * c)
+}
+
+// https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples
+fn next_triples(triplet: Triplet) -> List(Triplet) {
+  let Triplet(a, b, c) = triplet
+  [
+    Triplet(a - 2 * b + 2 * c, 2 * a - b + 2 * c, 2 * a - 2 * b + 3 * c),
+    Triplet(a + 2 * b + 2 * c, 2 * a + b + 2 * c, 2 * a + 2 * b + 3 * c),
+    Triplet(0 - a + 2 * b + 2 * c, -2 * a + b + 2 * c, -2 * a + 2 * b + 3 * c),
```
jiegillet, line 56: Here in the first term I initially wrote `-a + 2 * b + 2 * c` but the `-a` doesn't parse. It feels to me like it should work. Should I open a ticket?

exercises/practice/pythagorean-triplet/.meta/example.gleam
```diff
@@ -0,0 +1,64 @@
...
+        True -> [multiply_sides(triplet, sum / n), ..valid]
+        False -> valid
+      }
+
+      let candidates = case n >= sum {
+        True -> rest
+        False -> list.append(next_triples(triplet), rest)
+      }
+
+      generate_triples(candidates, sum, valid)
```
jiegillet, line 35: I tried to make it tail recursive, but actually I don't know if it's a thing in Gleam

jiegillet:
Foe example, here is a port of the first community solution of the Elixir track

```gleam
import gleam/list

pub type Triplet {
  Triplet(Int, Int, Int)
}

pub fn triplets_with_sum(sum: Int) -> List(Triplet) {
  list.range(1, sum)
  |> list.flat_map(fn(a) {
    list.range(a + 1, sum - a - 1)
    |> list.flat_map(fn(b) {
      let c = sum - a - b
      let triplet = Triplet(a, b, c)
      case c > b && pythagorean(triplet) {
        True -> [triplet]
        False -> []
      }
    })
  })
}

fn pythagorean(triplet: Triplet) -> Bool {
  let Triplet(a, b, c) = triplet
  a * a + b * b == c * c
}
```

With `gleam test`, the first 6 tests finish in 0.05s, but with the last one, it doesn't even finish

```gleam
➜  pythagorean-triplet git:(jie-pythagorean-triplet) ✗ gleam test
  Compiling pythagorean_triplet
   Compiled in 0.42s
    Running pythagorean_triplet_test.main
......
Pending:
  pythagorean_triplet_test.triplets_for_large_number_test
    %% Unknown error: {timeout,
                   #{stacktrace =>
   [...]
Finished in ? seconds
9 tests, 0 failures, 3 cancelled
```

exercises/practice/pythagorean-triplet/.meta/example.gleam
lpil: It is, yup!

jiegillet:
Ok, I've removed it for now. 
Thanks for the reviews!

lpil (approved):
Fab, thanks again!

PR #165, opened 2023-01-30
Add practice exercise `tournament`

jiegillet:
Neat one that's not too complex but involves a lot of different notions (strings, maps, ordering, option...).


exercises/practice/tournament/test/tournament_test.gleam
```diff
@@ -0,0 +1,153 @@
...
+  tournament.tally("")
+  |> should.equal("Team                           | MP |  W |  D |  L |  P")
+}
+
+pub fn a_win_is_three_points__a_loss_is_zero_points_test() {
+  tournament.tally("Allegoric Alaskans;Blithering Badgers;win")
+  |> should.equal(
+    "Team                           | MP |  W |  D |  L |  P
+Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3
+Blithering Badgers             |  1 |  0 |  0 |  1 |  0",
```
jiegillet, line 19: It bothers me a tiny bit that the header isn't aligned. 
I thought about adding a newline to align them and then passing it to `string.trim` to remove it but that seemed too noisy.

lpil: It is, isn't it. Annoys me all the time. Perhaps one day we can have something like Elixir's """

lpil (approved):
Nice! Thank you

PR #166, opened 2023-02-01
Add practice exercise `alphametics`

jiegillet:
This one was harder than I remembered, I wonder if there is a solution that doesn't involve brute force...

The last two test cases are the hardest, there is a chance that they would time out I suppose.
My Gleam solution takes 2.1 seconds, but the solution I wrote in Elixir a while back took 7.9 seconds. 
In this case though, I am inclined to leave them in, because both of these were brute force anyway.
What do you think?


exercises/practice/alphametics/.meta/example.gleam
```diff
@@ -0,0 +1,132 @@
...
+      from: map.new(),
+      with: fn(coefficients, term) { add_term(coefficients, term, int.add) },
+    )
+    |> add_term(right, int.subtract)
+
+  let leading_digits =
+    [right, ..terms]
+    |> list.map(string.slice(_, at_index: 0, length: 1))
+    |> list.unique
+    |> set.from_list
```
lpil, line 32: I think this goes `list -> set -> list -> set`, can skip the roundtrip
```suggestion
    |> set.from_list
```

lpil (approved):
Wow that one really does not look easy. Well done, thanks!

PR #167, opened 2023-02-03
Add practice exercise `robot-simulator`

jiegillet:
A nice simple one to get used to pattern matching.


exercises/practice/robot-simulator/.meta/example.gleam
```diff
@@ -0,0 +1,73 @@
...
+  West
+}
+
+pub type Position {
+  Position(x: Int, y: Int)
+}
+
+pub fn create(
+  direction direction: Direction,
+  position position: Position,
```
lpil, line 21: Is there a reason we've gone for labels with this one? I feel the tests would be a bit easier to read without them as the two arguments are clearly distinct so the labels are more noise than a help.

lpil (approved):
Very nice!

exercises/practice/robot-simulator/.meta/example.gleam
jiegillet: Sounds good. I'm sometimes on the fence when there are two arguments, but you are right, the arguments are really distinct here.

lpil (approved):
Thank you!

PR #170, opened 2023-02-04
Sync docs and filepaths

jiegillet:
I realized there was a slight bug in the generator script, configlet didn't include the `"invalidator"` files in the config since the files in questions were not created yet.

Also a missing `tests.toml` file (all tests were implemented already).

Also an instruction improvement pushed upstream by @michallepicki :)


lpil (approved):
Thank you!

PR #178, opened 2023-02-07
Added exercise for `Collatz conjecture`

suraj-mandal:
Hi! I have created another exercise for gleam. This time it is `collatz conjecture`. Please review it and let me know in case any changes are required!

Love to contribute [❤️](https://emojipedia.org/red-heart/).


config.json
```diff
@@ -591,6 +591,23 @@
...
         "difficulty": 3
+      },
+      {
+        "slug": "collatz-conjecture",
+        "name": "Collatz Conjecture",
+        "uuid": "576c4a50-6aef-4b28-a7b4-0e682004a690",
+        "practices": [
+          "int",
+          "maths",
+          "recursion"
```
jiegillet, line 602: I think it's best to keep the number of concepts to practice to one or two.
So maybe in this case we can drop maths? I think int got us covered for what we want to do.

exercises/practice/collatz-conjecture/.meta/example.gleam
```diff
@@ -0,0 +1,17 @@
...
+      case val % 2 == 0 {
+        True -> 1 + get_collatz(val / 2)
+        False -> 1 + get_collatz(3 * val + 1)
+      }
+  }
+}
+
+pub fn steps(number: Int) -> Result(Int, String) {
+  case number <= 0 {
+    True -> Error("Only positive numbers are allowed")
```
jiegillet, line 14: In general, we want to use custom types for the errors, it's better practice because you can patter match on it.
Something like `NonPositiveNumber`

jiegillet (commented):
Looks great, I only have a few minor comments :)

suraj-mandal:
Hi! Sorry for late response. I have updated the exercises and incorporated the changes as you have mentioned. Can you have a look into it?

exercises/practice/collatz-conjecture/.meta/example.gleam
```diff
@@ -0,0 +1,21 @@
+pub type NonPositiveNumber {
```
jiegillet, line 1: The type name should be more general, because usually we would have more than one sort of errors. 
```suggestion
pub type Error {
```

exercises/practice/collatz-conjecture/src/collatz_conjecture.gleam
```diff
@@ -0,0 +1,7 @@
+pub type NonPositiveNumber {
```
jiegillet, line 1:
```suggestion
pub type Error {
```

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.gleam
```diff
@@ -0,0 +1,37 @@
...
+}
+
+pub fn large_number_of_even_and_odd_steps_test() {
+  collatz_conjecture.steps(1_000_000)
+  |> should.equal(Ok(152))
+}
+
+pub fn zero_is_an_error_test() {
+  collatz_conjecture.steps(0)
+  |> should.equal(Error(collatz_conjecture.NonPositiveNumber))
```
jiegillet, line 31: While in general I prefer using qualified functions, I think it's OK to import the types and use them directly, it's more readable. But this is fine too, you don't have to change it if you prefer this.

jiegillet (approved):
Awesome, thank you!
I'm approving already, but I still left some small suggestions.

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.gleam
suraj-mandal: Can you elaborate on this? I kinda got confused

exercises/practice/collatz-conjecture/src/collatz_conjecture.gleam
suraj-mandal: Here, instead of creating the custom type Error, The Result type already has an Error type in it. Can't I use that only? I mean use Error(Nil) something like that?

If I use this strategy, then I will have to return something like Error(Error), since gleam is returning Result type object

jiegillet: No you will not return `Error(Error)`, you will return `Error(NonPositiveNumber)`, which will be of type `Result(a, Error)`. 
I've done that in other exercises, there will be no problem because `Error` is a constructor for the type `Result`, and `NonPositiveNumber` will be a constructor for the type `Error`. Since constructors and types are never used in the same context, there is no ambiguity so we can use the same name and the compiler will not be confused.

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.gleam
jiegillet: Sorry. let me try again ^^
This is not a big deal BTW, you don't really need to change anything if you don't want to.

What I mean by qualified function is
```gleam
import collatz_conjecture

fn something(){
  collatz_conjecture.steps(0)
}
```

as opposed to imported function
```gleam
import collatz_conjecture.{steps}

fn something(){
  steps(0)
}
```

In general, qualified functions are better because when you import many different libraries, there can be name clashes, functions become harder to track down, and it's nice to be able to associate unqualified names with functions defined in the same module.

For types however, I usually import them
```gleam
import collatz_conjecture.{NonPositiveNumber}

fn something(){
  Error(NonPositiveNumber)
}
```
because they are are usually a lot easier to identify since there are not that many.

suraj-mandal:
I have made the changes. Can you check?

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.gleam
```diff
@@ -0,0 +1,37 @@
+import gleeunit
+import gleeunit/should
+import collatz_conjecture.{Error, NonPositiveNumber}
```
jiegillet, line 3: You don't need the `Error` here, since it's not used. The compiler doesn't warn you about that? If not, it could be a bug...

```suggestion
import collatz_conjecture.{NonPositiveNumber}
```

suraj-mandal: Sorry, made a silly mistake.. fixing it ASAP!

jiegillet: You can just click "Commit suggestion" and it should work :)

jiegillet (approved):
🏅

PR #181, opened 2023-02-07
Added exercise for `Resistor color trio`

suraj-mandal:
Hi, I come again with addition of another exercise. This time it is `resistor-color-trio`.

Since this involves returning a custom data type, it was not being added in the gleam file the student will be using for submitting the solution. 

So to facilitate the student, the custom data type has been added, so the student is able to understand what exactly to return

### P.S.

Please review it and let me know if any changes are required. Will love feedback from your end.


exercises/practice/resistor-color-trio/.meta/example.gleam
```diff
@@ -0,0 +1,52 @@
+import gleam/map.{from_list, get}
+import gleam/int.{parse, to_string}
+import gleam/string.{repeat}
+import gleam/list.{at}
+
+pub type CustomRecordType {
+  CustomRecordType(unit: String, value: Int)
```
lpil, line 7: Let's give this a better name, "CustomRecordType" doesn't mean anything

exercises/practice/resistor-color-trio/.meta/example.gleam
```diff
@@ -0,0 +1,52 @@
...
+      #("green", 5),
+      #("blue", 6),
+      #("violet", 7),
+      #("grey", 8),
+      #("white", 9),
+    ])
+
+  assert Ok(first_color) = at(colors, 0)
+  assert Ok(second_color) = at(colors, 1)
+  assert Ok(multiplier) = at(colors, 2)
```
lpil, line 27: This would be best done with pattern matching as it's more idiomatic and it traversed once rather than 3 times.

```suggestion
  assert [first_color, second_color, multiplier, ..] = colors
```

exercises/practice/resistor-color-trio/.meta/example.gleam
```diff
@@ -0,0 +1,52 @@
...
+  let total_zeros = case second_color_val {
+    0 -> multiplier_val + 1
+    _ -> multiplier_val
+  }
+
+  case total_zeros {
+    9 -> CustomRecordType(unit: "gigaohms", value: value / 1_000_000_000)
+    6 | 7 | 8 -> CustomRecordType(unit: "megaohms", value: value / 1_000_000)
+    3 | 4 | 5 -> CustomRecordType(unit: "kiloohms", value: value / 1000)
+    0 | 1 | 2 -> CustomRecordType(unit: "ohms", value: value)
```
lpil, line 50: Let's have a catch all clause for when it is not one of these numbers.

lpil (commented):
Thank you!

It would be good to use imported functions in a qualified fashion, it would make the code easier to understand and it is idiomatic in Gleam to not use unqualified functions.
For example, `int.parse(value_string)` is easier to understand than `parse(value_string)`. Without the qualified I had to read the rest of the module to understand where the `parse` function comes from. I wrongly assumed it was a function defined in the module.

config.json
```diff
@@ -604,6 +604,26 @@
...
+          "custom-type",
+          "case"
+        ],
+        "prerequisites": [
+          "map",
+          "list",
+          "string",
+          "custom-type",
+          "case"
+        ],
```
lpil, line 625:
```suggestion
        "practices": [
          "maths",
          "list"
        ],
        "prerequisites": [
          "map",
          "list",
          "string",
          "custom-type",
          "case",
          "maths"
        ],
```

Perhaps this?

exercises/practice/resistor-color-trio/.meta/example.gleam
```diff
@@ -0,0 +1,52 @@
...
+      #("brown", 1),
+      #("red", 2),
+      #("orange", 3),
+      #("yellow", 4),
+      #("green", 5),
+      #("blue", 6),
+      #("violet", 7),
+      #("grey", 8),
+      #("white", 9),
+    ])
```
lpil, line 23: This never changes so would be best expressed using `case`. Maps are slower, use more memory, so there's no advantage if the mapping never changes.

suraj-mandal:
Thanks a lot for your reviews. I am fixing them ASAP. Happy to get a lot of insights!

suraj-mandal:
Hi! I have updated the exercise as suggested. Can you check and let me know if it is okay or not! Thanks!

lpil (approved):
Thank you!

PR #185, opened 2023-02-10
Added exercise called `Grains` to the gleam track

suraj-mandal:
Hi! I have created the PR for the given exercise called `Grains` to the Gleam track. Please review it and let me  know if any changes are required! Thanks!


exercises/practice/grains/test/grains_test.gleam
```diff
@@ -0,0 +1,65 @@
+import gleeunit
+import gleeunit/should
+import grains
+
+// The final tests of square test error conditions
+// In these cases you should expect an error as is idiomatic for your language
```
jiegillet, line 6: You can remove the comments

exercises/practice/grains/src/grains.gleam
```diff
@@ -0,0 +1,11 @@
+pub type InvalidSquare {
+  InvalidSquare
+}
+
+pub fn square(square square: Int) -> Result(Int, String) {
```
jiegillet, line 5:
```suggestion
pub fn square(square: Int) -> Result(Int, Error) {
```
(or whatever type you will end up having)

exercises/practice/grains/.meta/example.gleam
```diff
@@ -0,0 +1,21 @@
+pub type InvalidSquare {
```
jiegillet, line 1:
```suggestion
pub type Error {
```
(or some other name that is more general and can represent more possible errors)
Then change the type signature of `square`.

exercises/practice/grains/.meta/example.gleam
```diff
@@ -0,0 +1,21 @@
...
+  InvalidSquare
+}
+
+fn power(exponent: Int, result: Int) -> Int {
+  case exponent {
+    0 -> result
```
jiegillet, line 7: Shouldn't something raised to the 0th power be 1?
I'm sure the code works, but then calling it `power` is a bit misleading, because it makes me expect a generic power function. 
I would either implement a "real" power function and use that, or rename this one `grains_on_square` or something.

jiegillet (commented):
Looks quite nice already!

I really do appreciate all your contributions, but could I suggest that you submit only one or two PRs at a time? Because there are a lot of suggestions that are the same across  PRs.

suraj-mandal:
Sure.. i will do that.. thanks..

exercises/practice/grains/.meta/example.gleam
suraj-mandal: Actually you are correct, however, for the power function, I am tail recursion, so I am directly passing the value 1 as the ans beforehand. However, one thing here I am assuming is that this power function will be taking positive exponents only. This is valid in this scenario because the function only gets executed when the square is between 1 and 64 and not otherwise, so I have not bothered to add this condition since this function is a private function.

I have changed the name of the function because I agree with you, this power function is misleading.. I am changing it ASAP.

jiegillet (approved):
Well done, a fitting name makes all the difference :)

PR #187, opened 2023-02-12
run-length-encoding: add exercise

ErikSchierboom:


lpil (approved):
Oh wow that's a clever regex, very nice! I didn't realise you could do that.

ErikSchierboom:
Thanks! To be fair, I learnt it from a student :)

PR #189, opened 2023-02-12
high-scores: add exercise

ErikSchierboom:


lpil (commented):
Nice!

I'm not sure the API here is quite right for Gleam, it forces you write a partial function. I think `latest` and `personal_best` ought to return `Result(Int, Nil)` as there may be no scores in the list.

Tiny Gleam tip: Getting the last item from a list or reversing one is comparatively slow, in this case it would be more efficient to reverse the comparison function. We should add an `order.reverse` function to the stdlib to help with this!

ErikSchierboom:
I'll tweak it a bit!

exercises/practice/high-scores/.meta/example.gleam
```diff
@@ -0,0 +1,24 @@
...
+  high_scores
+}
+
+pub fn latest(high_scores: List(Int)) -> Result(Int, Nil) {
+  list.last(high_scores)
+}
+
+pub fn personal_best(high_scores: List(Int)) -> Result(Int, Nil) {
+  high_scores
+  |> list.sort(by: int.compare)
```
jiegillet, line 14: You could do this to avoid reversing
```suggestion
  |> list.sort(by: fn(a, b) { int.compare(b, a) })
```

config.json
```diff
@@ -755,6 +755,20 @@
         ],
         "difficulty": 2
       },
+      {
+        "slug": "high-scores",
+        "name": "High scores",
```
jiegillet, line 760:
```suggestion
        "name": "High Scores",
```

exercises/practice/high-scores/.meta/example.gleam
lpil: heh, that's actually more concise than the helper function I suggested :P

ErikSchierboom:
@jiegillet Any idea why it would fail in CI but not locally?

jiegillet:
This fails locally for me.
That's because I mislead you with my suggestion of changing the compare function, it was meant to replace the `list.reverse` line after it as well. It was my fault so I fixed it :)

ErikSchierboom:
Okay thanks!

PR #190, opened 2023-02-12
nucleotide-count: add exercise

ErikSchierboom:


exercises/practice/nucleotide-count/.meta/example.gleam
```diff
@@ -0,0 +1,23 @@
...
+import gleam/string
+import gleam/list
+import gleam/option.{None, Some}
+
+pub fn nucleotide_count(dna: String) -> Result(Map(String, Int), Nil) {
+  dna
+  |> string.to_graphemes()
+  |> list.try_fold(
+    map.from_list([#("A", 0), #("C", 0), #("G", 0), #("T", 0)]),
+    fn(counts, letter) {
```
jiegillet, line 11: Do you know about labels? They are pretty cool, I use them all the time with folds because I get confused with the order of arguments:
```gleam
|> list.try_fold(
  from: map.from_list([...])
  with: fn(counts, letter) { ...}
```

jiegillet (approved):
Awesome!
Thanks for all these contributions :)

I left a small comment just in case, but you don't need to act on it.

exercises/practice/nucleotide-count/.meta/example.gleam
ErikSchierboom: I've updated! I knew about labels, but hadn't yet come up with guidance for myself when to use them. Is there such a thing?

jiegillet: I doubt there is. My personal one is whenever there are 3 arguments :)

PR #192, opened 2023-02-13
proverb: add exercise

ErikSchierboom:


exercises/practice/proverb/test/proverb_test.gleam
```diff
@@ -0,0 +1,57 @@
+import gleeunit
+import gleeunit/should
+import proverb
+
+pub fn main() {
+  gleeunit.main()
+}
+
+pub fn zero_pieces_test() {
+  proverb.recite(inputs: [])
```
jiegillet, line 10: In this case I think we can remove the labels.
I think I should tweak the generator so that it doesn't use labels for a single argument. So far the hand waved rule has been: labels for 2 arguments with the same type or more than 2 arguments.

exercises/practice/proverb/test/proverb_test.gleam
```diff
@@ -0,0 +1,57 @@
...
+
+pub fn one_piece_test() {
+  proverb.recite(inputs: ["nail"])
+  |> should.equal(["And all for the want of a nail."])
+}
+
+pub fn two_pieces_test() {
+  proverb.recite(inputs: ["nail", "shoe"])
+  |> should.equal([
+    "For want of a nail the shoe was lost.", "And all for the want of a nail.",
```
jiegillet, line 22: I think here we should output a single multiline trimmed string.

```gleam
|> should.equal("For want of a nail the shoe was lost.
And all for the want of a nail.")
```

exercises/practice/proverb/.meta/example.gleam
```diff
@@ -0,0 +1,17 @@
+import gleam/list
+
+pub fn recite(inputs inputs: List(String)) -> List(String) {
+  case inputs {
+    [] -> []
+    [subject, ..] ->
+      inputs
+      |> list.window(2)
```
jiegillet, line 8: There is a specialized function for this `list.window_by_2` which is actually longer to type haha :D
The difference is that it produces pairs in tuples that are easier to pattern match on.

jiegillet (commented):
Very nice, I have a couple of suggestions

exercises/practice/proverb/.meta/example.gleam
```diff
@@ -0,0 +1,20 @@
...
+
+pub fn recite(inputs: List(String)) -> String {
+  case inputs {
+    [] -> ""
+    [subject, ..] ->
+      inputs
+      |> list.window_by_2()
+      |> list.map(fn(pair) {
+        case pair {
+          #(want, lost) ->
```
jiegillet, line 12: You can pattern match directly in the case of tuples and types with only one constructor:
```gleam
let #(want, lost) = pair
```
it will reduce the nesting.

exercises/practice/proverb/.meta/example.gleam
```diff
@@ -0,0 +1,20 @@
...
+            "For want of a " <> want <> " the " <> lost <> " was lost."
+        }
+      })
+      |> list.append(["And all for the want of a " <> subject <> "."])
+      |> list.intersperse("\n")
+      |> string.concat()
```
jiegillet, line 18:
```suggestion
      |> string.join("\n")
```

exercises/practice/proverb/.meta/example.gleam
```diff
@@ -0,0 +1,20 @@
...
+    [subject, ..] ->
+      inputs
+      |> list.window_by_2()
+      |> list.map(fn(pair) {
+        case pair {
+          #(want, lost) ->
+            "For want of a " <> want <> " the " <> lost <> " was lost."
+        }
+      })
+      |> list.append(["And all for the want of a " <> subject <> "."])
```
jiegillet, line 16: `list.append` is slow, maybe it's best to concatenate the final sentence to the string of all the other sentences. Not that it matters that much for 10 sentences :)

ErikSchierboom: I don't think it matters here, and what's more, if I were to append it I'd then have to conditionally add a newline before it, depending on whether the list is empty

exercises/practice/proverb/src/proverb.gleam
```diff
@@ -0,0 +1,17 @@
+import gleam/list
+import gleam/string
+
+pub fn recite(inputs: List(String)) -> String {
+  case inputs {
```
jiegillet, line 5: This one should be emptied out

jiegillet (approved):
One more thing to do, and then we are good to go!

exercises/practice/proverb/.meta/example.gleam
jiegillet: Yep, fair enough :)

exercises/practice/proverb/src/proverb.gleam
ErikSchierboom: Doh!

PR #193, opened 2023-02-13
etl: add exercise

ErikSchierboom:


exercises/practice/etl/.meta/example.gleam
```diff
@@ -0,0 +1,19 @@
+import gleam/string
+import gleam/map.{Map}
+import gleam/list
+
+pub fn transform(legacy legacy: Map(Int, List(String))) -> Map(String, Int) {
```
jiegillet, line 5: Let's remove the label here too!

jiegillet (approved):
Awesome sauce, thank you!
I left one comment, but I'm already approving :)

PR #194, opened 2023-02-13
allergies: add exercise

ErikSchierboom:


exercises/practice/allergies/gleam.toml
```diff
@@ -0,0 +1,9 @@
+name = "allergies"
+version = "0.1.0"
+
+[dependencies]
+gleam_stdlib = "~> 0.26"
+gleam_bitwise = "~> 1.2"
```
ErikSchierboom, line 6: I've relied on `gleam_bitwise` for this exercise. 
If we think that's okay, we'll need to ensure the test runner also supports this dependency.

jiegillet: This is a nice place to start this discussion.
I've had my eye on the `grep` exercise that requires to read files, and we would need the `gleam_erlang` dependency for that.

lpil: Adding `gleam_bitwise` sounds good. Unsure about `gleam_erlang` but we can figure that out later.

If we add additional packages do we need to do anything to communicate to the students what packages they can use?

ErikSchierboom: > If we add additional packages do we need to do anything to communicate to the students what packages they can use?

That's still an unsolved problem I'm afraid. We've discussed some options, but haven't yet implemented any.

exercises/practice/allergies/test/allergies_test.gleam
```diff
@@ -0,0 +1,282 @@
...
+import gleeunit/should
+import allergies
+
+pub fn main() {
+  gleeunit.main()
+}
+
+pub fn testing_for_eggs_allergy_not_allergic_to_anything_test() {
+  allergies.allergic_to(allergies.Eggs, 0)
+  |> should.equal(False)
```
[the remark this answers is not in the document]
ErikSchierboom: Okay cool!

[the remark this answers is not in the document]
ErikSchierboom: Updated

exercises/practice/allergies/gleam.toml
ErikSchierboom: See https://github.com/exercism/gleam-test-runner/pull/11

config.json
```diff
@@ -755,6 +755,23 @@
...
+          "int",
+          "list",
+          "bitwise"
+        ],
+        "prerequisites": [
+          "int",
+          "list",
+          "bitwise",
+          "case"
+        ],
```
lpil, line 772: This could also be a practice for bit strings as this could be solved using bit string pattern matching rather than bitwise operations. This is more common in Gleam/Erlang/etc

lpil (approved):
Nice!

config.json
ErikSchierboom: Yes, I'm quite unfamiliar with those. I've added `bit-string` to the `practices` list.

PR #197, opened 2023-02-13
rotational-cipher: add exercise

ErikSchierboom:


lpil (approved):
Oh that is very clever! Nice!

PR #198, opened 2023-02-13
Add practice exercise `prime-factors`

jiegillet:
It was still bugging me a bit that the factorization in perfect numbers was so slow, so... here goes a better way to do it :)
(I did not change perfect-numbers though)

I also tweaked the generator to only add labels when there are 2 arguments of the same type, or more than 3 arguments, which is more or less the rule we followed so far.


lpil (approved):
Fab. Thank you!

PR #199, opened 2023-02-13
Sync Gleam configuration

lpil:
Getting up to the latest version to match the test runner.


jiegillet:
Nice!
We have a number of PRs open that will be out of sync.
Should we take this opportunity to write a CI check that makes sure that all exercises have the same dependencies?
I did [something similar](https://github.com/exercism/elm/blob/e41f61f6c1ea2673fc2c7947621a63e5dded0cc2/bin/build.sh#L8-L12) on the Elm track (but the Elm template is in the repo, here we need to downloading from the test runner repo first).

lpil:
> Should we take this opportunity to write a CI check that makes sure that all exercises have the same dependencies?

We already have that :)

jiegillet:
Haha, indeed we do! Well done :D

jiegillet (approved):

PR #202, opened 2023-02-14
Add practice exercise `flatten-array`

jiegillet:
This one has a very simple solution, but I think can really showcase the beauty of functional programming with types :)

The original lists were something like `[1, 2, [3, [], null]]` so I made a custom type to model that.


suraj-mandal:
Hi! Thanks for the elegant solution, I created a similar thread in **exercism community forum** since I was having difficulty understand if it was at all possible to model such a list in gleam, since a list is containing atomic elements along with lists. From what I understand this concept was more aligned with dynamically typed language and thus was against the static type system of gleam. But your approach is very clever, and I got an idea of how to approach similar problem. Thanks!

jiegillet:
I'm glad you like the solution. I found the thread and MatthijsBlom's model for the nested list and solution is essentially the same as mine (minus `Nil`). He also mentioned that this exercise is not very well suited for Gleam because it's not really idiomatic, which I agree with, but would argue that it's still a good exercise to learn more about types, and Exercism is a place for learning :)

exercises/practice/flatten-array/.meta/example.gleam
```diff
@@ -0,0 +1,15 @@
+import gleam/list
+
+pub type NestedList(a) {
+  Nil
```
lpil, line 4: Let's call this `Null` to match the spec and to not shadow the prelude `Nil`.

exercises/practice/flatten-array/.meta/example.gleam
```diff
@@ -0,0 +1,15 @@
...
+  Value(a)
+  NestedList(List(NestedList(a)))
```
lpil, line 6: How about just `List` for this value constructor? It's quite long as it's not always nested, it could be at the top level, matching the other constructors also.

jiegillet: Ok, I hope the `  List(List(NestedList(a)))` is not too confusing :)

lpil (approved):
Wonderful! Thank you

PR #203, opened 2023-02-14
Sync manifests after adding `gleam_bitwise`

jiegillet:
I was a bit surprised to not see any CI failures, turns out there was a bug in the logic.
This commit is supposed to fail the CI, because there are unsynced manifests in the repo.
I will fix them later, then the CI should pass.

### DIFF
--- a/bin/sync-exercise-project-configs
+++ b/bin/sync-exercise-project-configs
@@ -31,8 +31,8 @@ check() {
   project="$1"
   update_config_name "$project"
 
-  if ! diff "$canonical_config" "$project"/gleam.toml && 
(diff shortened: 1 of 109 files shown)


jiegillet:
[495f571](https://github.com/exercism/gleam/pull/203/commits/495f57154b2208641744da409e63fd0589bcbf34) failed and [f00ece2](https://github.com/exercism/gleam/pull/203/commits/f00ece28712db7eede50ab29157b4aa4d9404a8a) passed, as expected.

lpil (approved):
Thank you

PR #204, opened 2023-02-14
Proposal: alternative way to run tests

jiegillet:
I got a little bit impatient looking at the tests run, so I though I would propose this alternate approach that we use on the Elm track of copying all solutions and tests into one big project and run that.

This takes the process a bit further away from what the students would experience, so I understand if you want to turn this down.

This version runs in 14s, as opposed to almost 4 minutes for the current version on my machine.

The output in case of an error is decent:

```
➜  gleam git:(main) ✗ bin/test
Copying accumulate practice exercise...
Copying [...many more...]
Copying zipper practice exercise...
  Compiling all_exercises
   Compiled in 1.83s
    Running all_exercises_test.main
.....F..................................................................................................................................................................................
.........................................................................................................................................................................................
.........................................................................................................................................................................................
...................................................................................................................................................................
Failures:

  1) acronym_test.basic_test
     Values were not equal
     expected: "PNG - I tweaked this test"
          got: "PNG"
     output: 

Finished in 9.310 seconds
717 tests, 1 failures
A solution did not pass the tests
```

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -74,7 +74,7 @@ function copy_exercise {
 }
 
 function run_all {
-  # create a all_exercises folder
+  # create an all_exercises folder
   mkdir -p all_exercises all_exercises/src all_exercises/test
 
   cat <<EOT > all_exercises/test/all_exercises_test.gleam
@@ -111,7 +111,7 @@ EOT
   done
 
   # Running all tests
-    if ! (cd all_exercises && gleam test)
+  if ! (cd all_exercises && gleam test)
   then
     echo "A solution did not pass the tests"
     exit_code=1
--- a/bin/test
+++ b/bin/test
@@ -44,26 +44,80 @@ function verify_exercise {
   mv "${stub_backup_file}" "${stub_file}"
 }
 
+download() {
+  file="$1"
+  url="$2"
+  curl --silent --show-error --fail --retry 3 --max-time 3 \
+    --output "$file" "$url"
+}
+
+function copy_exercise {
+  local dir slug implementation_file stub_backup_file module
+  dir="$1"
+  dest="$2"
+  slug=$(basename "${dir}")
+  module=${slug//-/_}
+
+  implementation_destination="${dest}/src/${module}.gleam"
+  practice_solution="${dir}/.meta/example.gleam"
+  if [ -f "${practice_solution}" ]
+   then cp "${practice_solution}" "${implementation_destination}"
+  fi
+  concept_solution="${dir}/.meta/src/${module}.gleam"
+  if [ -f "${concept_solution}" ]
+   then cp "${concept_solution}" "${implementation_destination}"
+  fi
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -44,26 +44,80 @@ function verify_exercise {
...
     then
-      echo "Checking $(basename "${practice_exercise_dir}") practice exercise..."
-      verify_exercise "${practice_exercise_dir}"
+      echo "Copying $(basename "${practice_exercise_dir}") practice exercise..."
+      copy_exercise "${practice_exercise_dir}" all_exercises
     fi
   done
+
+  # Running all tests
+    if ! (cd all_exercises && gleam test)
```
lpil, line 114:
```suggestion
  if ! (cd all_exercises && gleam test)
```

bin/test
```diff
@@ -44,26 +44,80 @@ function verify_exercise {
...
+  fi
+
+  test_file="${dir}/test/${module}_test.gleam"
+  test_destination="${dest}/test/${module}_test.gleam"
+  cp "${test_file}" "${test_destination}"
+}
+
 function run_all {
-  # Verify the Concept Exercises
+  # create a all_exercises folder
```
lpil, line 77:
```suggestion
  # create an all_exercises folder
```

lpil (approved):
Hey cool! Very nice :)

If we wanted to make it even faster we could use the JavaScript backend which doesn't have to run the Erlang compiler.

jiegillet:
How do we do that? I've never tried switching backend

lpil:
You can either set `target = "javascript"` in `gleam.toml` or give the `--target=javascript` flag.

jiegillet:
Ok, I will leave it like this, it's not like 14s is very long.

I tried the JS flag anyway and got 
```
Downloading packages
 Downloaded 3 packages in 0.04s
  Compiling gleam_stdlib
  Compiling gleam_bitwise
  Compiling gleeunit
  Compiling all_exercises
error: Fatal compiler bug!

This is a bug in the Gleam compiler, sorry!

Please report this crash to https://github.com/gleam-lang/gleam/issues/new
with this information and the code that produces the crash.

        compiler-core/src/javascript/pattern.rs:403
        argument present in field map
A solution did not pass the tests
```
hahaha.

I tried the JS backend on each exercise and:
- `armstrong-numbers` fails because it has test cases with very large integers (up to 127 bits)
- `prime-factors` fails, for the same reason
- `pythagorean-triplet` doesn't compile
- `pov`doesn't compile

Should I open issues? :)

For `pov`, I've narrowed it down to
```gleam
type Zipper(a) {
  Zipper(focus: a, List(a))
}

fn to_rerouted_tree(zipper: Zipper(a)) -> a {
  let Zipper(tree, crumbs) = zipper
  tree
}
```
This fails with the error above.

PR #208, opened 2023-02-15
Added exercise for `All Your Base`

suraj-mandal:
Hello, I have created an exercise for the problem `all your base`. Please review it and let me know in case any changes are required or not.

I might need some help in the naming. Unable to find a better name for the error and functions in the meantime.

Thanks!


exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,45 @@
...
+  digits digits: List(Int),
+  input_base input_base: Int,
+  output_base output_base: Int,
+) -> Result(List(Int), Error) {
+  case input_base, output_base {
+    input, output if input >= 2 && output >= 2 ->
+      case get_base_number(digits, input) {
+        Ok(0) -> Ok([0])
+        Ok(value) -> Ok(get_base_digits(value, output, []))
+        _ -> Error(InvalidDigit)
```
jiegillet, line 19: Here maybe I would write this instead
```suggestion
        error -> error
```
Because there might be other types of errors going on, you don't want to override them.

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,45 @@
...
+      }
+    _, _ -> Error(InvalidBase)
+  }
+}
+
+fn get_base_number(digits: List(Int), base: Int) -> Result(Int, Error) {
+  digits
+  |> list.fold(
```
jiegillet, line 27: In Elixir, it's usually advised not to use a pipe if there is only one function applied. When I see a pipe, I quickly glance at all the steps to get an idea of the data transformations before reading the details, and it feels strange when there is no next step. 
Also, not using the pipe reduces the nesting a little bit.

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,45 @@
...
+    Ok(0),
+    fn(acc, cur) {
```
jiegillet, line 29: 3 letter variable names are really hard to read, how about something like
`acc` -> `result`
`cur` -> `digit`

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,45 @@
...
+      case cur, acc {
+        _, Error(InvalidDigit) -> Error(InvalidDigit)
+        value, acc if value >= 0 && value < base ->
+          result.then(acc, fn(acc_val) { Ok(acc_val * base + value) })
```
jiegillet, line 33: I think it would be easier to read to pattern match `acc` with `Ok(value)` so you can use `value` directly.

Another possible approach is using `result.map`, then you probably don't need the pattern match on the errors. Mixing the two is strange.

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,45 @@
...
+        _, _ -> Error(InvalidDigit)
+      }
+    },
+  )
+}
+
+fn get_base_digits(value: Int, base: Int, result_list: List(Int)) -> List(Int) {
+  case value {
+    0 -> result_list
+    _ -> get_base_digits(value / base, base, [value % base, ..result_list])
```
jiegillet, line 43: Love this ❤️

jiegillet (commented):
Really nice implementation, let's make it more readable so that it shines in all its glory ✨💖✨

exercises/practice/all-your-base/.meta/example.gleam
suraj-mandal: gotcha, trying this..

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,46 @@
+import gleam/list
+import gleam/result
+
+pub type Error {
+  InvalidBase(Int)
+  InvalidDigitSequence(Int)
```
jiegillet, line 6: I really love this, but now the name is a bit wrong, since we are carrying the invalid digit and presenting it, it should be named `InvalidDigit`. `InvalidDigitSequence` would be a better name to mean "your sequence is invalid for some unspecified reason".

exercises/practice/all-your-base/src/all_your_base.gleam
```diff
@@ -0,0 +1,12 @@
+pub type Error {
+  InvalidBase(Int)
+  InvalidDigit(Int)
```
jiegillet, line 3: This value is inconsistent with the new solution, but since it's what it should be, no change needed.

exercises/practice/all-your-base/test/all_your_base_test.gleam
```diff
@@ -0,0 +1,131 @@
...
+  all_your_base.rebase(digits: [1], input_base: 2, output_base: -7)
+  |> should.equal(Error(InvalidBase(-7)))
+}
+
+pub fn both_bases_are_negative_test() {
+  let output =
+    all_your_base.rebase(digits: [1], input_base: -2, output_base: -7)
+  should.be_true(
+    output == Error(InvalidBase(-2)) || output == Error(InvalidBase(-7)),
+  )
```
jiegillet, line 130: I didn't see this coming, nicely handled. Let's use `assert` though
```suggestion
  assert True = output == Error(InvalidBase(-2)) || output == Error(InvalidBase(-7)),
```
(this probably need formatting)

jiegillet: Or maybe we could just do
```gleam
assert Error(InvalidBase(_)) = output
```
Whichever you prefer

exercises/practice/all-your-base/.meta/example.gleam
```diff
@@ -0,0 +1,46 @@
...
+fn get_base_number(digits: List(Int), base: Int) -> Result(Int, Error) {
+  list.fold(
+    digits,
+    Ok(0),
+    fn(accumulator, digit) {
+      case digit, accumulator {
+        value, Ok(current) if value >= 0 && value < base ->
+          Ok(current * base + value)
+        value, present_val ->
+          result.then(present_val, fn(_) { Error(InvalidDigitSequence(value)) })
```
jiegillet, line 35: I get why this works and the `result.then` is necessary instead of returning the error directly, but it feels more complicated than it could be. Have you ever used `list.fold_until`? I think it could help

jiegillet (commented):
Very nice, I left some more comments.
I hope I'm not being too nitpicky, your solution is already great!

suraj-mandal:
Making the changes.

exercises/practice/all-your-base/.meta/example.gleam
suraj-mandal: I am checking the list.fold_until?, however the purpose of result.then was that if already the error was there then it would return the error, or it would create the error since we know that now the value is not satisfy the condition.

suraj-mandal:
Hi! I have made the changes, sorry for making mistakes all the time. Please review it once!

exercises/practice/all-your-base/.meta/example.gleam
jiegillet: Yes, like I said, I understand how it works, I just thought we could make it a bit less complicated.
Another benefit of `list.fold_until` is that you stop checking all the digits as soon as you found an invalid one.

jiegillet (approved):
This is great, thank you!

And please, do not apologize for making small mistakes, it's completely normal, expected even, and that's why code reviews are important. I really value your contributions.

Also, I know that I can be very nitpicky sometimes when I believe the code can be improved with small efforts, if I go too far in that direction, feel free to let me know and I will back out :)

PR #209, opened 2023-02-15
yacht: add exercise

ErikSchierboom:


exercises/practice/yacht/.meta/example.gleam
```diff
@@ -0,0 +1,91 @@
...
+    [_] -> 50
+    _ -> 0
+  }
+}
+
+fn dice_with_count(dice: List(Int)) -> List(#(Int, Int)) {
+  dice
+  |> list.group(function.identity)
+  |> map.to_list()
+  |> list.map(fn(dice_count) { pair.map_second(dice_count, list.length) })
```
jiegillet, line 89: I think it would be a tiny (tiny) bit more readable to use `map.map_values` before turning the map into a list because `map.map_values` separates the key and value, so you wouldn't need to deal with a tuple.

jiegillet (approved):
Beautiful solution!
I left a small comment that you are free to ignore if it's too nitpicky :D

exercises/practice/yacht/.meta/example.gleam
ErikSchierboom: I originally had that. Not even sure why I changed it. Updated!

ErikSchierboom (commented):
I like niticks BTW! Keep them coming

jiegillet:
@ErikSchierboom Oh no! You updated the stub file instead of the example file haha :D

ErikSchierboom:
Oh no! 😱

ErikSchierboom:
https://github.com/exercism/gleam/pull/211

PR #210, opened 2023-02-16
Add practice exercise `pov`

jiegillet:
This one is pretty hard, but I'm quite happy with my solution, hopefully it's readable :)


config.json
```diff
@@ -463,6 +463,23 @@
         ],
         "difficulty": 6
       },
+      {
+        "slug": "pov",
+        "name": "POV",
+        "uuid": "2012322c-1693-4702-abfe-1f3e7b176035",
+        "practices": [
+          "custom-type"
```
lpil, line 471: Perhaps this also practices result?

lpil (approved):
What a good looking solution! Nice

config.json
jiegillet: Sounds good

PR #213, opened 2023-02-17
dnd-character: add exercise

ErikSchierboom:


config.json
```diff
@@ -940,6 +940,21 @@
...
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "65d80b55-02f0-4d40-a530-79843553252f",
+        "practices": [
+          "int",
+          "random",
```
lpil, line 950: Correct me if I'm wrong, but shouldn't all the practices be in the prerequisites also? Have to have learnt it to practice it?

lpil (approved):
Looks excellent. A fun exercise too :)

I've left 1 comment inline

config.json
ErikSchierboom: Yep, my mistake.

config.json
```diff
@@ -940,6 +940,21 @@
...
+          "custom-type"
+        ],
+        "prerequisites": [
+          "int",
+          "custom-type"
```
ErikSchierboom, line 955:
```suggestion
          "random",
          "custom-type"
```

PR #219, opened 2023-02-18
atbash-cipher: add exercise

ErikSchierboom:


exercises/practice/atbash-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,32 @@
...
+import gleam/map.{Map}
+
+pub fn encode(phrase: String) -> String {
+  let encode_map = encoding_map()
+
+  phrase
+  |> string.lowercase()
+  |> string.to_graphemes()
+  |> list.map(fn(char) { map.get(encode_map, char) })
+  |> result.values()
```
jiegillet, line 13: `result.values` is so valuable :)

config.json
```diff
@@ -983,6 +983,18 @@
...
+      },
+      {
+        "slug": "atbash-cipher",
+        "name": "Atbash Cipher",
+        "uuid": "3c850415-92fa-4019-b776-aaab30675680",
+        "practices": [
+          "string"
+        ],
+        "prerequisites": [
+          "string"
```
jiegillet, line 995: You also used list, result, and map, let's add that.
Also maybe difficulty is a 3? not sure

jiegillet (commented):
Really clean solution, thank you!

exercises/practice/atbash-cipher/.meta/example.gleam
ErikSchierboom: Absolutely!

config.json
ErikSchierboom: I find the difficulty hard to gauge.

> You also used list, result, and map, let's add that.

Well, that I used them in my implementation doesn't mean that people _have_ to use them. 
Ideally, the prerequisites should only list the minimal set of concepts a student has to know. If one person solves the exercise via a map but another doesn't use a map, then `map` should not really be a prerequisite as you don't need to know about the `map` concept to solve the exercise.
That said, `list` is something people have to work with, due to `strings.to_graphemes()`, so I added that.

jiegillet: Well, it's a grey area, but I think the prerequisite can also hint at a clean way to solve the exercise. I would have to go out of my way to find a decent solution that doesn't use maps, but maybe I lack imagination :)
I don't mind too much though, `list` was the most important in my suggestions.

And +1 on the difficulty being hard to gauge. Although it doesn't matter that much. 
I keep thinking that we could re-calibrate the exercise difficulty for tracks using some data after the exercise has been on for a while, maybe the percentage of people who tried the exercise but did not finish or something like that.

jiegillet (approved):

config.json
ErikSchierboom: > I would have to go out of my way to find a decent solution that doesn't use maps, but maybe I lack imagination :)

Well, I'll trust your judgment then and will add `map` to the list of prerequisites.

> maybe the percentage of people who tried the exercise but did not finish or something like that.

Yeah, that will be really useful. https://exercism.org/tracks/gleam/build can help with that

PR #220, opened 2023-02-18
space-age: add exercise

ErikSchierboom:


exercises/practice/space-age/.meta/tests.toml
```diff
@@ -0,0 +1,38 @@
...
+
+[999354c1-76f8-4bb5-a672-f317b6436743]
+description = "age on Uranus"
+
+[80096d30-a0d4-4449-903e-a381178355d8]
+description = "age on Neptune"
+
+[57b96e2a-1178-40b7-b34d-f3c9c34e4bf4]
+description = "invalid planet causes error"
+include = false
```
jiegillet, line 38: Let's add a comment for posterity, something like
```
comment = "planets are encoded as types and cannot be invalid"
```

exercises/practice/space-age/.meta/example.gleam
```diff
@@ -0,0 +1,32 @@
...
+  Venus
+  Earth
+  Mars
+  Jupiter
+  Saturn
+  Uranus
+  Neptune
+}
+
+pub fn age(planet: Planet, seconds: Int) -> Float {
```
jiegillet, line 15: What would you think of `seconds` to take a float? It would make sense from an API perspective.

jiegillet (commented):
Thank you, it looks great!
I left some comments

exercises/practice/space-age/.meta/tests.toml
ErikSchierboom: Good idea

exercises/practice/space-age/.meta/example.gleam
ErikSchierboom: I had considered that and thought that maybe converting from an int to a float would be useful practice, but maybe not 🤷 I updated to a float.

jiegillet (approved):
Perfect, thank you!

PR #221, opened 2023-02-19
Add practice exercise `kindergarten-garden`

jiegillet:
Pretty straightforward one, uses a decent amount of `list` functions, although `string.slice` approaches would work fine as well.


lpil (approved):
Very nice!

PR #223, opened 2023-02-19
Add practice exercise `strain`

natanaelsirqueira:
First time authoring an exercise so picked a small one. :) 

There are no tests in the problem specification for this exercise so I copied the ones used in the Elixir track. Should these be added to the problem specification or is there a reason not to?


jiegillet:
Thank you for the contribution! 
You solution looks great, but I must admit I would feel a lot more comfortable if the test cases were in the problem-specifications, it really makes a big difference for maintainability.

It would be perfect if we could add them to the specs before merging this PR. If you are willing to help with this, you could start a thread on the forum, then I would be happy to champion your efforts and support you. Recently, a [PR has been opened](https://github.com/exercism/problem-specifications/pull/2192) to do the same with another exercise, so I think it would be well received.
What do you think?

lpil:
Great idea!

natanaelsirqueira:
@jiegillet Thanks. Raised the discussion [here](https://forum.exercism.org/t/proposal-to-add-test-cases-to-the-strain-problem-spec/3961).

natanaelsirqueira:
@lpil I tried to apply what you did [here](https://github.com/exercism/gleam/pull/252) but the CI check is still failing. What am I doing wrong?

lpil:
I think one of the other configs may be wrong on your branch? Does running `./bin/sync-exercise-project-configs` help?

natanaelsirqueira:
The problem specs PR has been merged! 🙌  

I already revised the tests here to match the canonical data.

lpil (approved):
Fab!! Thank you 💜

PR #224, opened 2023-02-20
affine-cipher: add exercise

ErikSchierboom:


exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,84 @@
...
+      let r = a % b
+      let #(s, t, g) = gcd_ext(b, r)
+      #(t, s - q * t, g)
+    }
+  }
+}
+
+const alphabet_length = 26
+
+const min_int_code_point = 97
```
jiegillet, line 84: Maybe here a comment would be nice, specifying what 97 is.

config.json
```diff
@@ -1045,6 +1045,24 @@
...
+      },
+      {
+        "slug": "affine-cipher",
+        "name": "Affine Cipher",
+        "uuid": "594b6c1a-0a4c-4463-9535-62056c71404e",
+        "practices": [
+          "string",
+          "list",
+          "int",
+          "result"
```
jiegillet, line 1057: Let's narrow down that list a bit.
Maybe string and int? Or string and math? I think math makes sense here.

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,84 @@
...
+import gleam/list
+import gleam/int
+
+pub fn encode(
+  plaintext plaintext: String,
+  a a: Int,
+  b b: Int,
+) -> Result(String, Nil) {
+  modular_inverse(a, alphabet_length)
+  |> result.map(fn(_) {
```
jiegillet, line 12: I think a `case` would be a bit more readable than passing a value into a pipe only to ignore it immediately :)

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,84 @@
...
+
+pub fn decode(
+  ciphertext ciphertext: String,
+  a a: Int,
+  b b: Int,
+) -> Result(String, Nil) {
+  modular_inverse(a, alphabet_length)
+  |> result.map(fn(mmi) {
+    ciphertext
+    |> translate(fn(index) { mmi * { index - b } })
```
jiegillet, line 30: I usually avoid pipes when there is only one function being applied. A pipe sets an expectation of several steps, and it feels strange to only have one.

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,84 @@
...
+        string.utf_codepoint(min_int_code_point + shifted_index)
+      }
+      False ->
+        case list.contains(digits, char) {
+          True -> Ok(char)
+          False -> Error(Nil)
+        }
+    }
+  })
+  |> result.values()
```
jiegillet, line 59: You can use `list.filter_map` so you don't need this line

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,84 @@
...
+  })
+}
+
+pub fn decode(
+  ciphertext ciphertext: String,
+  a a: Int,
+  b b: Int,
+) -> Result(String, Nil) {
+  modular_inverse(a, alphabet_length)
+  |> result.map(fn(mmi) {
```
jiegillet, line 28: I'm always on the lookout for a use of [`use`](https://gleam.run/news/v0.25-introducing-use-expressions/).
I think this could be a perfect spot. Or at least `try`

exercises/practice/affine-cipher/.meta/example.gleam
ErikSchierboom: I've tried a naming refactoring. See if that works for you>?

exercises/practice/affine-cipher/.meta/example.gleam
lpil: This is also the `result.replace` function if you don't mind it being eager.

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,89 @@
...
+  |> string.lowercase()
+  |> string.to_utf_codepoints()
+  |> list.filter_map(fn(char) {
+    case list.contains(letters, char) {
+      True -> {
+        assert Ok(shifted_index) =
+          int.modulo(
+            op(string.utf_codepoint_to_int(char) - letter_a_int_code_point),
+            alphabet_length,
+          )
```
lpil, line 54: In this case you could use the `%` operator and not need an assert. `int.modulo` is for when you care if the right hand side is 0 or not.

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,89 @@
...
+}
+
+pub fn encode(
+  plaintext plaintext: String,
+  a a: Int,
+  b b: Int,
+) -> Result(String, Error) {
+  try _ =
+    modular_inverse(a, alphabet_length)
+    |> result.replace_error(KeyNotCoprime(a, b))
```
jiegillet, line 17: You don't need to replace the error, just have `modular_inverse` return the correct error.
Here `a` and `alphabet_length` are the problem, so that's what should be returned, the instructions are pretty clear about this, so it wouldn't be a surprise to the students.

exercises/practice/affine-cipher/.meta/example.gleam
```diff
@@ -0,0 +1,89 @@
...
+}
+
+pub fn decode(
+  ciphertext ciphertext: String,
+  a a: Int,
+  b b: Int,
+) -> Result(String, Error) {
+  try mmi =
+    modular_inverse(a, alphabet_length)
+    |> result.replace_error(KeyNotCoprime(a, b))
```
jiegillet, line 35: Same here

exercises/practice/affine-cipher/.meta/example.gleam
ErikSchierboom: You're right. I've updated

exercises/practice/affine-cipher/.meta/example.gleam
ErikSchierboom: I can't unfortunately, as the `%` doesn't do the "wraparound" that I need to correctly handle negative values

jiegillet (approved):
Very nice!!

PR #226, opened 2023-02-20
largest-series-product: add exercise

ErikSchierboom:


lpil (approved):
Very nice! Good use of the result module

PR #227, opened 2023-02-20
secret-handshake: add exercise

ErikSchierboom:


config.json
```diff
@@ -1045,6 +1045,24 @@
...
         "difficulty": 5
+      },
+      {
+        "slug": "secret-handshake",
+        "name": "Secret Handshake",
+        "uuid": "22461f35-684d-4f75-9399-d07a4f483bdd",
+        "practices": [
+          "int",
+          "list",
+          "bitwise",
```
lpil, line 1056: This may also be solved with bit string pattern matching in BEAM languages, so perhaps we can put bitstring in here also.

lpil (approved):
Very nice!

PR #229, opened 2023-02-20
clock: add exercise

ErikSchierboom:


exercises/practice/clock/src/clock.gleam
```diff
@@ -0,0 +1,17 @@
+// TODO: define a custom 'Clock' type
```
ErikSchierboom, line 1: I thought it would maybe be nice for a student to have to define the Clock type themselves

lpil: Perhaps we could give them this? It would be a syntax error

```gleam
pub type Clock {
}
```

exercises/practice/clock/.meta/example.gleam
```diff
@@ -0,0 +1,39 @@
+import gleam/string
+import gleam/int
+
+pub type Clock {
+  Clock(minutes: Int)
+}
+
+pub fn create(hour hour: Int, minute minute: Int) -> Clock {
+  assert Ok(minutes) =
+    int.modulo(hour * minutes_per_hour + minute, minutes_per_day)
```
lpil, line 10: Any reason not to use the `%` operator here? Looks like you don't care about the error case of `int.modulo`

lpil (approved):
Nice!!

exercises/practice/clock/src/clock.gleam
jiegillet: I usually go with
```gleam
pub type Clock {
  TODO
}
```

exercises/practice/clock/.meta/example.gleam
ErikSchierboom: There is a specific reason. `%` doesn't do "wraparound", so for negative values it doesn't give me the answer I want (whereas `int.modulo` does).

exercises/practice/clock/src/clock.gleam
ErikSchierboom: Updated

PR #233, opened 2023-02-22
Add practice exercise `twelve-days`

natanaelsirqueira:


exercises/practice/twelve-days/test/twelve_days_test.gleam
```diff
@@ -0,0 +1,143 @@
...
+  should.equal(
+    verse(12),
+    "On the twelfth day of Christmas my true love gave to me: twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+  )
+}
+
+pub fn first_three_verses_of_the_song_test() {
+  should.equal(
+    lyrics(from: 1, to: 3),
+    string.join(
```
jiegillet, line 97: Let's usea single  multiline string here:
```gleam
    "On the first day of Christmas my true love gave to me: a Partridge in a Pear Tree.
On the second day of Christmas my true love gave to me: two Turtle Doves, and a Partridge in a Pear Tree."
```

exercises/practice/twelve-days/.meta/example.gleam
```diff
@@ -0,0 +1,51 @@
+import gleam/list
+import gleam/pair
+import gleam/string
+
+const day_gift_pairs = [
```
jiegillet, line 5: I think making this into a function with a `case`
```gleam
case number {
  1 ->   #("first", "a Partridge in a Pear Tree")
  2 -> ...
}
```
could potentially simplify `verse` a bit.

exercises/practice/twelve-days/.meta/example.gleam
```diff
@@ -0,0 +1,51 @@
...
+    _ -> {
+      let [this_verse_day_gift_pair, ..previous_verses_day_gift_pairs] =
+        remaining_day_gift_pairs
+        |> list.take(number - 1)
+        |> list.reverse()
+
+      pair.map_second(
+        this_verse_day_gift_pair,
+        fn(gift) {
+          [gift, ..list.map(previous_verses_day_gift_pairs, pair.second)]
```
jiegillet, line 35: This is bit hard to follow.
Maybe we could extract the day and the gifts in separate steps? Something like (untested)
```gleam
      let gift_pairs =
        remaining_day_gift_pairs
        |> list.take(number - 1)
        |> list.reverse()
      
      assert [#(day, _), ..] = gift_pairs

      let gifts = gift_pairs
        |> list.map(pair.second)
        |> string.join(", ")
        |> string.append(", and " <> first_day_gift)

      #(day, gifts)
```

config.json
```diff
@@ -834,6 +834,21 @@
...
       },
+      {
+        "slug": "twelve-days",
+        "name": "Twelve Days",
+        "uuid": "583b74dc-2c5c-48f8-85c2-c90623507682",
+        "difficulty": 3,
+        "practices": [
+          "string",
+          "case",
+          "list"
```
jiegillet, line 845: Whatever is practiced should be a prerequisite, also it's better to limit this list to one or two elements. 
Maybe we can drop `case`?

jiegillet (commented):
Very nice solution, thank you!
I left some comments for things we can improve further :)

config.json
natanaelsirqueira: One could solve this without using lists (by writing `verse` with just case and `lyrics` with recursion). It wouldn't look great, but it's possible. 

That said, should `list` be dropped instead or should it be kept because we want people to use `list` for this exercise? (I'm still trying to understand what drives the decision of what goes into `practices`/`prerequisites`).

natanaelsirqueira:
@jiegillet Thanks for the feedback! I started learning Gleam this month so it's nice to receive some hints on how to make my code look better. :)

exercises/practice/twelve-days/.meta/example.gleam
```diff
@@ -0,0 +1,44 @@
...
+import gleam/string
+
+pub fn verse(number: Int) -> String {
+  let #(day, gift) = day_gift_pair_for_verse(number)
+
+  let gifts = case number {
+    1 -> gift
+    _ ->
+      number
+      |> list.range(2)
```
jiegillet, line 12: It's good practice to start a pipe with a raw value, but I always make an exception for `list.range`, because `list.range(number, 2)` is way more readable.

config.json
jiegillet: It's not an exact thing, but `prerequisites` should show concepts that are needed for an idiomatic solution, that's why looking at what was necessary in the solution example is a good indication. For now, Gleam doesn't have any concepts, so this list isn't actually used, but once it does, the exercises will be blocked until the prerequisite concepts are unlocked.
So if we omit `list`, students that have not yet been taught about lists may start this exercise. I think they would have a harder time.

The `practices` should have one or two of the most important concepts. They will be suggested to the students as soon as they unlock the concept in question. That's why I would remove `case`, it's likely to be a more basic concept that comes before lists and string, so it couldn't actually be recommended.

Of course, there can always be solutions that don't use all of those concepts, and that's fine.

jiegillet (commented):
Awesome, your solution looks even better now, and I won't refrain from nitpicks from now on, in the name of learning lol 😈

I left one more tiny suggestion, and we need to fix the config, then it'll be good to merge.

config.json
natanaelsirqueira: Interesting... that makes things a lot clearer for me, thanks!

exercises/practice/twelve-days/.meta/example.gleam
natanaelsirqueira: Makes sense. I also thought it looked nicer when I used labels for creating the range but went with the convention you mentioned which I picked from Elixir. 😅

jiegillet: Yeah, I could tell, I'm also an Elixir dev :D

PR #234, opened 2023-02-22
roman-numerals: specify return type

ErikSchierboom:
Not sure if this (not defining the return type) was intentional or not.


jiegillet (approved):
Intentional or not, this is an improvement, thank you!

PR #235, opened 2023-02-23
Add practice exercise `connect`

jiegillet:
This one deals with graph traversal on a hexagonal lattice, a bit harder than average.


config.json
```diff
@@ -1080,7 +1080,7 @@
         "uuid": "594b6c1a-0a4c-4463-9535-62056c71404e",
         "practices": [
           "string",
-          "math"
+          "maths"
```
jiegillet, line 1083: I changed this for consistency, we had both math and maths. I picked math because it had more occurrences 🤷

lpil: I've gone for the English spelling everywhere so maths is good by me 👍

exercises/practice/connect/.meta/example.gleam
```diff
@@ -0,0 +1,152 @@
+import gleam/list
+import gleam/map.{Map}
+import gleam/string
+import gleam/result
+
+pub type Player {
+  X
+  O
+}
```
lpil, line 9: Not sure why but I find this type definition so satisfying

exercises/practice/connect/.meta/example.gleam
```diff
@@ -0,0 +1,152 @@
...
+  Size(rows: Int, columns: Int)
+}
+
+pub fn winner(board: String) -> Result(Player, Nil) {
+  let rows =
+    board
+    |> string.trim()
+    |> string.split("\n")
+
+  assert [first_row, ..] = rows
```
lpil, line 25: I would return an error here, the board could be empty.

lpil (approved):
Very nice! Looks like quite a tricky one.

I've left one small note

PR #236, opened 2023-02-23
Add practice exercise `say`

jiegillet:
My solution for this one might not be the most clever or streamlined, but I realized halfway in that I don't actually like this one that much, so it'll do :D


lpil (approved):
Looks perfectly good to me!

PR #241, opened 2023-02-25
Add practice exercise `minesweeper`

jiegillet:
New one, pretty standard :)


exercises/practice/minesweeper/.meta/tests.toml
```diff
@@ -0,0 +1,48 @@
...
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[0c5ec4bd-dea7-4138-8651-1203e1cb9f44]
+description = "no rows"
+
+[650ac4c0-ad6b-4b41-acde-e4ea5852c3b8]
+description = "no columns"
+include = false
+comment = "empty board already covered by 0c5ec4bd-dea7-4138-8651-1203e1cb9f44"
```
jiegillet, line 18: This test was transforming `"\n"` into `"\n"`, but in my implementation, `"\n"` is considered an empty board that outputs an empty string. I felt that this was an unnecessary edge case to consider.

exercises/practice/minesweeper/test/minesweeper_test.gleam
```diff
@@ -0,0 +1,144 @@
...
+pub fn space_surrounded_by_mines_test() {
+  minesweeper.annotate(
+    "***
+*_*
+***",
+  )
+  |> should.equal(
+    "***
+*8*
+***",
```
jiegillet, line 62: This doesn't look super nice, but introducing a leading newline in the output would be weird.

lpil (approved):
What a fun exercise! Thank you

PR #242, opened 2023-02-25
nth-prime: add exercise

ErikSchierboom:


config.json
```diff
@@ -1197,6 +1197,19 @@
...
         ],
         "difficulty": 2
+      },
+      {
+        "slug": "nth-prime",
+        "name": "Nth Prime",
+        "uuid": "fda0001e-51b5-4041-a08a-2cb7dfef0f1c",
+        "practices": [
+          "int",
+          "iterator"
```
ErikSchierboom, line 1207: Should this also be a prerequisite? I don't know how people would implement it without it, but maybe using plain recursion or list filtering?

lpil: I think normal recursion would be quite normal

lpil (approved):
Nice!!

PR #243, opened 2023-02-25
Add practice exercise `triangle`

natanaelsirqueira:
Supporting both ints & floats would be a bit tricky and increase the difficulty of the exercise so I chose not to include floats. Is this a fair assumption to make?

There are a few other options we could consider:

1. Leave the three tests that use floats as optional (commented) in the test file and create a second example solution that handles ints & floats - just in case the student is up for the challenge.
2. Create an advanced version of this exercise that requires supporting ints & floats and can be used to teach `gleam/dynamic` (similar to what was done with [Take-a-Number](https://exercism.org/tracks/elixir/exercises/take-a-number) and [Take-a-Number Deluxe](https://exercism.org/tracks/elixir/exercises/take-a-number-deluxe) in the Elixir track).
3. Turn this exercise into that advanced version and increase its difficulty.

I'm more inclined towards leaving this exercise simple and _maybe_ doing option 2 later. What do you reckon?


exercises/practice/triangle/.meta/tests.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[c6585b7d-a8c0-4ad8-8a34-e21d36f7ad87]
+description = "equilateral triangle -> no sides are equal"
+
+[16e8ceb0-eadb-46d1-b892-c50327479251]
+description = "equilateral triangle -> all zero sides is not a triangle"
+
+[3022f537-b8e5-4cc1-8f12-fd775827a00c]
+description = "equilateral triangle -> sides may be floats"
+include = false
```
lpil, line 26: How about the sides are always floats? That way we can include all the tests

exercises/practice/triangle/.meta/example.gleam
```diff
@@ -0,0 +1,36 @@
+import gleam/int
+import gleam/list
+
+pub fn equilateral(sides: List(Int)) -> Bool {
```
lpil, line 4: The code looks great!

The type `List(Int)` implies that the shape could have any number of sides, or even none. In Gleam we use the type system to enforce properties in the program, so here I think we would want to enforce that triangles have 3 sides.

We could do this by having the functions take 3 arguments, or making a `Triangle(Float, Float, Float)` record to represent a triangle.

lpil (commented):
Nice! Left a couple comments inline

exercises/practice/triangle/.meta/tests.toml
natanaelsirqueira: That sounds sensible!

exercises/practice/triangle/.meta/example.gleam
natanaelsirqueira: Thanks for the input! Went with separate arguments.

exercises/practice/triangle/.meta/example.gleam
```diff
@@ -0,0 +1,36 @@
...
+  let is_valid_triangle = is_valid_triangle(a, b, c)
+
+  case list.unique([a, b, c]) {
+    [_, _, _] if is_valid_triangle -> True
+    _ -> False
+  }
+}
+
+fn is_valid_triangle(a: Float, b: Float, c: Float) -> Bool {
+  let [a, b, c] = list.sort([a, b, c], float.compare)
```
lpil, line 33: This pattern isn't exhaustive so it should use `assert` rather than `let`!

natanaelsirqueira: Fixed!

lpil (approved):
Wonderful work! Thank you!

PR #244, opened 2023-02-26
Add practice exercise `wordy`

natanaelsirqueira:
Wanted to have some fun so picked a more complex one. :)


config.json
```diff
@@ -1178,6 +1178,25 @@
...
+          "result",
+          "string",
+          "case",
+          "list",
+          "bool",
+          "int",
+          "maths",
+          "custom-type"
+        ],
+        "difficulty": 8
```
natanaelsirqueira, line 1198: This exercise has difficulty 7 in Elixir. I gave it one level more here because it's a bit tricky to handle the different edge cases and you must be comfortable with the type system and a few different concepts.

exercises/practice/wordy/test/wordy_test.gleam
```diff
@@ -0,0 +1,127 @@
...
+  wordy.answer("What is 1 2 plus?")
+  |> should.equal(Error(SyntaxError))
+}
+
+pub fn reject_prefix_notation_test() {
+  wordy.answer("What is plus 1 2?")
+  |> should.equal(Error(SyntaxError))
+}
+
+pub fn reject_division_by_zero_test() {
```
natanaelsirqueira, line 124: I had to adapt the solution a bit to normalize the operations since `int.divide` is the only one that returns a `Result`, so I figured I'd add this extra test case for the sake of it.

exercises/practice/wordy/.meta/example.gleam
```diff
@@ -0,0 +1,135 @@
...
+
+fn evaluate(tokens: List(Token)) -> Result(Int, Error) {
+  try [Operand(result)] =
+    list.try_fold(
+      tokens,
+      [],
+      fn(previous_tokens: List(Token), current_token: Token) {
+        case previous_tokens {
+          [Operator(operation), Operand(left_operand)] -> {
+            assert Operand(right_operand) = current_token
```
lpil, line 55: Could be nice to return an error here rather than crashing, especially as this function already is a `Result` returning function.

lpil (approved):
Oh this is very nice! I left one request and a small suggestion in the example. Great stuff!

exercises/practice/wordy/.meta/example.gleam
natanaelsirqueira: Nice. I did this because the sequence of tokens was already being validated at the tokenization phase, so I thought it would be redundant to check for an error here. Now I realize that it makes more sense to do it here and leave the `parse_tokens` function cleaner since this validation is only relevant to this particular implementation of the `evaluate` function.

I like this solution better now. :)

lpil (approved):
Oh I **really** like this new version, fantastic!

PR #247, opened 2023-02-28
Add practice exercise `luhn`

natanaelsirqueira:
Simple and fun. :)


exercises/practice/luhn/.meta/example.gleam
```diff
@@ -0,0 +1,57 @@
+import gleam/int
+import gleam/list
+import gleam/result
+import gleam/string
+
+pub fn valid(value: String) -> Bool {
+  case
```
jiegillet, line 7: It would be a bit more readable to put the result in a variable before using `case`

jiegillet: Or actually, you might not need a case at all if you do
```gleam
    value
    |> parse_digits()
    |> result.then(validate_length)
    |> result.map(is_valid_number)
    |> result.is_ok
```

jiegillet (commented):
Really nice solution!
I left a couple of suggestions ^^

exercises/practice/luhn/test/luhn_test.gleam
```diff
@@ -0,0 +1,117 @@
...
+import gleeunit/should
+import luhn
+
+pub fn main() {
+  gleeunit.main()
+}
+
+pub fn single_digit_strings_can_not_be_valid_test() {
+  luhn.valid("1")
+  |> should.equal(False)
```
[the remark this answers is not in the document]
natanaelsirqueira: Thanks for the heads-up!

exercises/practice/luhn/.meta/example.gleam
natanaelsirqueira: Cool! It needs to be `result.unwrap(False)` instead of `result.is_ok()` though, or it would yield `True` regardless of the result of `is_valid_number`. :)

jiegillet: Yes, you're right! Nice :)

jiegillet (approved):
Thank you, that's great :)

PR #248, opened 2023-02-28
Add practice exercise `rational-numbers`

natanaelsirqueira:


exercises/practice/rational-numbers/.meta/example.gleam
```diff
@@ -0,0 +1,111 @@
...
+      RationalNumber(
+        numerator: int.negate(numerator),
+        denominator: int.negate(denominator),
+      )
+
+    numerator, denominator -> RationalNumber(numerator, denominator)
+  }
+}
+
+fn power_of_integer(base: Int, to exponent: Int) -> Result(Int, Nil) {
```
jiegillet, line 94: It's a bit strange to have labels on some arguments but not others, I could have sworn the compiler would complain...

exercises/practice/rational-numbers/.meta/example.gleam
```diff
@@ -0,0 +1,111 @@
...
+
+      Ok(power)
+    }
+  }
+}
+
+pub fn power_of_real(
+  number base: Int,
+  to exponent: RationalNumber,
+) -> Result(Float, Nil) {
```
jiegillet, line 70: It feels strange to have a `Result` here but no test with an error case...
What do you think? 
Maybe we should add a test case with a negative base to the problem-specs?

exercises/practice/rational-numbers/.meta/example.gleam
```diff
@@ -0,0 +1,111 @@
...
+        to: int.absolute_value(exponent),
+      )
+
+    RationalNumber(numerator, denominator) -> {
+      try numerator = power_of_integer(numerator, to: exponent)
+      try denominator = power_of_integer(denominator, to: exponent)
+
+      let power = reduce(RationalNumber(numerator, denominator))
+
+      Ok(power)
```
jiegillet, line 62: I don't think a rational to an integer power can actually fail, can it?

exercises/practice/rational-numbers/.meta/example.gleam
```diff
@@ -0,0 +1,111 @@
...
+
+  reduce(RationalNumber(numerator: a * d, denominator: c * b))
+}
+
+pub fn absolute_value(r: RationalNumber) -> RationalNumber {
+  let RationalNumber(numerator, denominator) = reduce(r)
+
+  RationalNumber(
+    numerator: int.absolute_value(numerator),
+    denominator: int.absolute_value(denominator),
```
jiegillet, line 41: Isn't the denominator always positive after reducing?

jiegillet (commented):
Nice work! Let's iron out some wrinkles :)

exercises/practice/rational-numbers/.meta/example.gleam
natanaelsirqueira: So... I did this because `power_of_integer` can fail (because `int.power` can fail). Should we make it so it doesn't return a `Result`?

```gleam
fn power_of_integer(number base: Int, to exponent: Int) -> Int {
  assert Ok(power) = int.power(base, int.to_float(exponent))

  float.round(power)
}
```

We could also just use `assert` on lines 57 and 58 instead:

```gleam
    RationalNumber(numerator, denominator) -> {
      assert Ok(numerator) = power_of_integer(numerator, to: exponent)
      assert Ok(denominator) = power_of_integer(denominator, to: exponent)

      reduce(RationalNumber(numerator, denominator))
    }
```

exercises/practice/rational-numbers/.meta/example.gleam
natanaelsirqueira: Sounds good. I'll work on a PR later today.

natanaelsirqueira: Opened the [PR](https://github.com/exercism/problem-specifications/pull/2213) and it got closed as expected. Are you able to reopen?

jiegillet: Thank you. Let's wait for the PR to be merged before merging this one :)

exercises/practice/rational-numbers/.meta/example.gleam
jiegillet: `int.power` can fail when the exponent is fractional, but we know that it is integer and positive here, so we can safely use `assert`. We even skip `power_of_integer` entirely and use `int.power`, but that's a matter of style.

PR #249, opened 2023-03-01
grade-school: add exercise

ErikSchierboom:


lpil (approved):
Very nice! Thank you

I did wonder if the roster should be a set rather than a list, but there's not much in it.

ErikSchierboom:
> I did wonder if the roster should be a set rather than a list, but there's not much in it.

Well it could be, but then we could no longer have the ordering test cases which are in the canonical data. That's not to say we need to conform to that, but it does make this slightly more involved (sort first by grade, then by name).

lpil (approved):
Makes sense, let's keep it as a list. Thanks

PR #253, opened 2023-03-04
Add practice exercise `variable-length-quantity`

jiegillet:
@lpil you will like this one, it's great to practice bitstrings.


exercises/practice/variable-length-quantity/.meta/example.gleam
```diff
@@ -0,0 +1,46 @@
...
+}
+
+fn encode_one(integer: Int, is_first_call: Bool) -> BitString {
+  let leading_bit = case is_first_call {
+    True -> 0
+    False -> 1
+  }
+
+  let small_bits = integer % 128
+  let small_byte = <<leading_bit:1, small_bits:7>>
```
jiegillet, line 22: Here I first tried
`<<leading_bit:1, integer % 128:7>>`
but it didn't work. Then I tried
`<<leading_bit:1, {integer % 128} :7>>`
but the `{}` got stripped away by the formatter.

Is this worth opening an issue or is it already on your radar?

lpil: Could you open an issue please? 🙏

lpil (approved):
Very nice!!

exercises/practice/variable-length-quantity/.meta/example.gleam
jiegillet: Done! https://github.com/gleam-lang/gleam/issues/2045

PR #254, opened 2023-03-04
Add exercise `resistor-color-duo`

natanaelsirqueira:


exercises/practice/resistor-color-duo/.meta/example.gleam
```diff
@@ -0,0 +1,33 @@
...
+  Yellow
+  Green
+  Blue
+  Violet
+  Grey
+  White
+}
+
+pub fn value(colors: List(Color)) -> Int {
+  let [color1, color2, ..] = colors
```
lpil, line 15: This would crash if the list is less than 2 elements long, and in future it will be rejected by the compiler.

How about we modify this function to return a `Result`, returning `Error(Nil)` or some other error when there's fewer than 2 elements?

lpil (commented):
Nice work, thank you! I've left a comment inline

exercises/practice/resistor-color-duo/.meta/example.gleam
natanaelsirqueira: That makes sense! Making the change...

exercises/practice/resistor-color-duo/.meta/example.gleam
```diff
@@ -0,0 +1,34 @@
...
+  Blue
+  Violet
+  Grey
+  White
+}
+
+pub fn value(colors: List(Color)) -> Result(Int, String) {
+  case colors {
+    [color1, color2, ..] -> Ok(10 * color_code(color1) + color_code(color2))
+    _ -> Error("the list must contain at least two colors")
```
lpil, line 17: Let's use `Nil` or a custom error type rather than strings. String errors are discouraged in Gleam as they encode information but what the variants are is much less clear than with a custom type.

natanaelsirqueira: Duly noted. Thanks!

lpil (approved):
Thank you

PR #255, opened 2023-03-04
Add practice exercise `saddle-points`

jiegillet:
New one.
I wonder whether we should add a comment about the positions being 1-indexed, it follows the description but might be a source of friction.


lpil (approved):
Nice! Thanks!!

exercises/practice/saddle-points/test/saddle_points_test.gleam
```diff
@@ -0,0 +1,60 @@
...
+import gleeunit/should
+import saddle_points.{Position}
+
+pub fn main() {
+  gleeunit.main()
+}
+
+pub fn can_identify_single_saddle_point_test() {
+  saddle_points.saddle_points([[9, 8, 7], [5, 3, 2], [6, 6, 7]])
+  |> should.equal([Position(row: 2, column: 1)])
```
natanaelsirqueira, line 11: Curious to know when is it okay to use `should.equal` instead of `assert expected = ...`

Does [this](https://github.com/exercism/gleam/issues/148#issuecomment-1400020512) apply only to boolean assertions?

jiegillet: I'm going to say yes, at the moment it only applies to `should.be_true`, although this will change in the future, because the behavior of `assert` is planned to evolve.

The reason I'm saying yes now is the quality of error messages. If a test fails with `assert`, you get a message like `Assertion pattern match failed. Value: (value that didn't match). Location: (file:line)`.
When you fail with a `should.equal`, you get: `Values were not equal. Expected: (expected value). Got: (value that didn't match)`.

One gives you a line number, the other one gives you both values so you can compare them directly, so I much prefer `should.equal` when I'm solving an exercise. For a boolean though, if it's not `True`, it's `False`, so the line number is more valuable because you can look up the failing test if you need to.

Other tracks (like Elixir) have test case parsers, so they always show you the code snippet of failing tests, in that case it wouldn't matter that much, but Gleam doesn't have that yet (ideally we would need a Gleam language parser in Gleam).

PR #257, opened 2023-03-07
Add practice exercise `ocr-numbers`

jiegillet:
The instructions mention the output should be a string, I thought a custom type would be more idiomatic, but I can switch it back...


lpil (approved):
Very nice! From reading the task I was expecting it to be more difficult but your solution makes it look easy :)

PR #259, opened 2023-03-08
Bump Gleam stdlib to 0.27

fapdash:
I saw that you bumped the Gleam version in CI but didn't bump the stdlib. I tried out locally if this would break any tests and it looks fine.

### DIFF
--- a/exercise_generator/gleam.toml
+++ b/exercise_generator/gleam.toml
@@ -3,7 +3,7 @@ version = "0.1.0"
 description = "Test and stub file generator for the Exercism Gleam track"
 
 [dependencies]
-gleam_stdlib = "~> 0.26"
+gleam_stdlib = "~> 0.27"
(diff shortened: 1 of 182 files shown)


lpil:
Hello! Thank you for your help here, but in this instance it has already been done. The constraint there permits v0.27, and v0.27 specifically is locked-in by the manifest file, like is like npm's package lock file.

https://github.com/exercism/gleam/blob/62537ceed5bfdfb3ba32fae3bb5814f4be49af29/exercise_generator/manifest.toml#L7

Next next time an update is due to change has to be made in the [runner project](https://github.com/exercism/gleam-test-runner) and then the [sync script](https://github.com/exercism/gleam/blob/main/bin/sync-exercise-project-configs) can be run in this project to sync them all.

fapdash:
Ah, very cool, thanks for the info :)

PR #260, opened 2023-03-11
Add exercise `custom-set`

natanaelsirqueira:
I put this one right before Circular Buffer as they are kind of similar.


exercises/practice/custom-set/src/custom_set.gleam
```diff
@@ -0,0 +1,49 @@
+pub opaque type CustomSet(t) {
+  CustomSet
```
lpil, line 2: How about we just call it `Set`?

lpil (commented):
Great solution, really nice. Just left one small question

exercises/practice/custom-set/src/custom_set.gleam
natanaelsirqueira: Done!

lpil (approved):
Brilliant! Thank you

PR #269, opened 2023-04-11
sync sum of multiples docs

kytrinyx:
- Sync sum-of-multiples docs with problem-specifications
- Sync sum-of-multiples docs with problem-specifications


kytrinyx:
My bad! Sorry about that. It should be good now.

lpil (approved):
Thank you

PR #278, opened 2023-05-08
Basics concept

lpil:


exercises/concept/lasagna/.docs/instructions.md
```diff
@@ -0,0 +1,50 @@
...
+Define the `total_time_in_minutes` function that takes two arguments: the first argument is the number of layers you added to the lasagna, and the second argument is the number of minutes the lasagna has been in the oven. The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
+
+```gleam
+total_time_in_minutes(3, 20)
+// -> 26
+```
+
+## 5. Create a notification that the lasagna is ready
+
+Define the `alarm/0` function that does not take any arguments and returns a message indicating that the lasagna is ready to eat.
```
hayleigh-dot-work: > Define the **`alarm/0`** function ...

Oops, sneaking in a bit of erlang syntax there!

exercises/concept/lasagna/.meta/design.md
```diff
@@ -0,0 +1,42 @@
+# Design
+
+## Learning objectives
+
+- Know what a variable is.
```
hayleigh-dot-work, line 5: This might sound obtuse but because you explicitly list this separate from "know how to update/define...", you don't actually ever explain what a variable is.

exercises/concept/lasagna/.docs/instructions.md
giacomocavalieri, line 45: Why is `alarm` the only function where you specify the `/0`? I think it might be confusing for a newcomer (also I thought it was an Erlang thing, I never noticed it in Gleam)

lpil: Well spotted, copy paste fail on my part.

lpil:
> The contents of concepts/basics/about.md, concepts/basics/introduction.md, and exercises/concept/lasagna/.docs/introduction.md are all the same, is that right?

Yup, that's right. We can add more info to `about.md` if we want, but it doesn't get shown much so Erik said not to worry about it much initially.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -0,0 +1,64 @@
...
+```gleam
+pub fn main() {
+  let count = 1
+
+  // You may shadow existing variables with the same name
+  let count = "Lots"
+}
+```
+
+Variables can be declared with type annotations. Like function arguments these are optional, though most Gleam programmers will omit type annotations for variables.
```
hayleigh-dot-work, line 47: > though most Gleam programmers will omit type annotations for variables.

Can we fit in an explanation why? For functions we say that Gleam programmers add the types for clarity, it might be a bit confusing to be told the opposite is the case for variables with no explanation

hayleigh-dot-work (approved):
LGTM, nothing deal-breaking.

exercises/concept/lasagna/.meta/design.md
lpil: Good point! Concepts are aimed at people who already know how to program, so I think it's OK not to teach that. I'll change the wording here.

exercises/concept/lasagna/.docs/introduction.md
lpil: For sure. "most Gleam programmers will omit type annotations for variables for brevity." ?

PR #280, opened 2023-05-08
Case expressions

lpil:
Follows on from #279


concepts/case-expressions/introduction.md
```diff
@@ -0,0 +1,54 @@
...
+    0 -> "Zero"
+    i if i < 0 -> "Negative number"
+    _ -> "Positive number"
+  }
+}
+
+// No compiler error
+```
+
+Case expressions will test a value against each pattern from top to bottom, until it finds a matching pattern and executes the logic associated with that pattern. The order of patterns matters!
```
giacomocavalieri, line 54: Are unreachable branches out of scope for this lesson? Otherwise it could be useful to show one example with an unreachable branch due to the order of patterns; I was thinking something like:
```gleam
case x {
  _ -> "Not Zero"
  0 -> "Zero"
}
```

concepts/case-expressions/introduction.md
```diff
@@ -0,0 +1,54 @@
...
+```gleam
+pub fn describe(number: Int) -> String {
+  case number {
+    0 -> "Zero"
+    1 -> "One"
+  }
+}
+```
+
+While this may look like `switch` statements in other languages, pattern matching starts to shine when also using other patterns. One such pattern is the _variable pattern_, which allows one to capture a value:
```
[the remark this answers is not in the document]
lpil: Good call!

concepts/case-expressions/introduction.md
lpil: I think the general idea is to only introduce what is needed to be known, rather than cover everything. Perhaps this would be a good addition to the `about` document, which can have more. I'll do that

PR #284, opened 2023-05-09
Use unique UUIDs in config.json

ErikSchierboom:
Two of the new UUIDs have the same value as their Elixir counterparts, which causes track syncing to fail.

Closes https://github.com/exercism/gleam/issues/283
Closes https://github.com/exercism/gleam/issues/281


lpil (approved):
Ah! Thank you.

PR #286, opened 2023-05-12
change custom_set tests to use is_equal

bcpeinhardt:
We also have a test that verifies the is_equal functionality so this should be fine


lpil (commented):
There's no assertion now! Will need to add `assert True = ` to the start of each

bcpeinhardt:
Whoops!!! Haha, will do when I can!

bcpeinhardt:
Added the `let assert True` bits!

lpil (approved):
Thank you 💜

PR #287, opened 2023-05-15
Tuples concept

lpil:


concepts/tuples/about.md
```diff
@@ -0,0 +1,65 @@
+# Introduction
+
+A tuple is an ordered container of values. Like all Gleam data types tuples are immutable.
+Each element of a type can be of a different type -- they can even be other tuples.
```
schurhammer, line 4: Each element of a ~~type~~ tuple

lpil: Thank you

PR #294, opened 2023-05-27
Anonymous functions concept

lpil:


exercises/concept/secrets/.docs/instructions.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## 5. Create a function combiner
+
+Implement `secret_combine`. It should return a function which takes one argument and applies to it the two functions passed in to `secret_combine` in order.
+
+```gleam
+let multiply = secret_multiply(7)
+let divide = secret_divide(3)
+let combined = secret_combine(multiply, divide)
```
giacomocavalieri, line 54: I don't remember if in one of the previous lessons type annotations and in particular the type annotation of functions were explained. If not, a beginner might find some difficulties trying to write the `secret_combine` function:
```gleam
pub fn secret_combine(f: ???, g: ???) -> Int {
  todo
}
```
If not already explained it would be difficult to come up with the types of `f` and `g`

exercises/concept/secrets/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+fn add_one(x) {
+  x + 1
+}
+```
+
+Gleam also has anonymous functions, which are defined within other functions.
+
+```gleam
+// Define the function
+let f = fn(x) { x + 1 }
```
giacomocavalieri, line 28: Connecting to my previous comment, maybe it could be useful to show just once the type annotation of such function, maybe in a comment:
```gleam
// f : fn(Int) -> Int
let f = fn(x) { x + 1 }
```

exercises/concept/secrets/.meta/design.md
```diff
@@ -0,0 +1,23 @@
...
+After completing this exercise, the student should:
+
+- Know how to write anonymous functions.
+- Know how to return a function from a function.
+- Know that anonymous functions can reference variables in scope when they were defined.
+- Know about the function capture syntax.
+
+## Out of scope
+
+- Higher order function patterns.
```
giacomocavalieri, line 14: Decide wether knowing the type signature of functions can be a learning objective or out of scope

exercises/concept/secrets/.docs/instructions.md
lpil: The type annotation is already in the code outline the user has to complete, but yup including the type annotation would be good!

giacomocavalieri: Oops sorry I completely missed the code outline!

PR #295, opened 2023-05-30
Sync tests

### LINKED ISSUE #134: Add practice exercise missing test cases

giacomocavalieri:
I ran `bin/configlet sync --tests --update` to update any dated test. Moreover, I looked for all occurrences of `include = false` in the codebase and either implemented the missing test or added a reason why the test is missing:
- any test that is reimplemented by another now explains that that is the reason why it is not included
- all tests that were meant to rule out mutable state were ignored with a proper reason
- all other tests were not hard to implement so I added them all

Since now there is no missing test and all tests that are not included were properly commented this should close #134 

Edit: there is one exception that is the Forth case: there are 3 tests that were previously ignored that the example doesn't pass. This would require fixing the forth example, I think this PR is already big enough so I've opened a new issue and will start working on that

### DIFF
--- a/exercises/practice/forth/test/forth_test.gleam
+++ b/exercises/practice/forth/test/forth_test.gleam
@@ -231,13 +231,13 @@ pub fn calling_a_nonexistent_word_test() {
   |> error_with(forth.UnknownWord)
 }
 
-pub fn user_defined_words_execute_in_the_right_order_test() {
-  forth.new()
-  |> forth.eval(": countup 1 2 3 ;")
-  |> result.then(forth.eval(_, "countup"))
-  |> result.map(forth.format_stack)
-  |> succeed_with("1 2 3")
-}
+//pub fn user_defined_words_execute_in_the_right_order_test() {
+//  forth.new()
+//  |> forth.eval(": countup 1 2 3 ;")
+//  |> result.then(forth.eval(_, "countup"))
+//  |> result.map(forth.format_stack)
+//  |> succeed_with("1 2 3")
+//}
 
 pub fn user_defined_words_can_override_builtin_operators_test() {
   forth.new()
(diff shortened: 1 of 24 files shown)


exercises/practice/forth/test/forth_test.gleam
```diff
@@ -158,3 +230,88 @@ pub fn calling_a_nonexistent_word_test() {
...
   |> error_with(forth.UnknownWord)
 }
+
+//pub fn user_defined_words_execute_in_the_right_order_test() {
+//  forth.new()
+//  |> forth.eval(": countup 1 2 3 ;")
+//  |> result.then(forth.eval(_, "countup"))
+//  |> result.map(forth.format_stack)
+//  |> succeed_with("1 2 3")
+//}
```
lpil, line 240: What are these comments for?

lpil (commented):
Thank you! I've left one comment

exercises/practice/forth/test/forth_test.gleam
giacomocavalieri: It's the tests that the current example for `forth` cannot pass because it needs a little fix. Maybe I should remove them altogether from this PR and add them back when I have a fix for Forth, what do you think?

lpil: Either works for me. If keeping them could you add a comment explaining what they are for please?

lpil (approved):
Thank you 💜

PR #299, opened 2023-06-05
Generics concept

lpil:


concepts/generics/introduction.md
```diff
@@ -0,0 +1,40 @@
...
+Type parameters can also be used in the arguments of functions.
+
+This function takes a value of the type `a` and calls twice a function that takes an `a` and returns a `b`.
+
+```gleam
+pub fn twice(value: a, f: fn(a) -> b) -> b {
+  f(value)
+  f(value)
+}
+```
```
giacomocavalieri, line 40: Maybe we could also show an example here:

```gleam
twice(1, fn(x) { int.to_string(x) }) // -> Has type fn(Int, fn(Int) -> String) -> String
```

giacomocavalieri (approved):
Look great to me! I left just a comment

PR #316, opened 2023-07-28
Orders and constants

lpil:


concepts/constants/about.md
```diff
@@ -0,0 +1,70 @@
+# About
```
giacomocavalieri, line 1: Why is this file in the constant's about? It doesn't look like it's about constants?

concepts/orders/about.md
```diff
@@ -0,0 +1,13 @@
+# About
```
giacomocavalieri, line 1: Why is this file under the orders' about when it is about constants?

lpil: Oh bums

lpil: Phew, the previous version was in the git history, that's a relief

exercises/concept/weather-ranking/.docs/introduction.md
```diff
@@ -0,0 +1,85 @@
+# Introduction
```
giacomocavalieri, line 1: Everything looks good to me now! The only thing I'm not sure about is why is this file used as the introduction for the `weather_ranking` exercise since it is about compare and constants?

lpil: Oops, thank you. I forgot to regenerate this file

PR #320, opened 2023-08-07
Labelled arguments

lpil:


giacomocavalieri (approved):
Looks great to me! Just curious to know, what is "anonymous function degrading" that is referenced in the out-of-scope section?

lpil:
If you assign a function with labels to a variable it degrades to an anonymous function and the labels can no longer be used.

giacomocavalieri:
> If you assign a function with labels to a variable it degrades to an anonymous function and the labels can no longer be used.

Ooh I didn't know that, thanks!

PR #323, opened 2023-08-11
let-assert

lpil:


exercises/concept/spring-cleaning/.docs/instructions.md
```diff
@@ -0,0 +1,37 @@
+# Instructions
+
+Mika the events organise has got a promotion at work, and as part of her new responsibilities she is in charge of the events database. She quickly noticed that a lot of the data is stored in ways that make it hard to work with, so you're teaming up to clean it up with a one-off script.
```
giacomocavalieri, line 3: > Mika _the events organise_ has got a promotion at work

There's something that doesn't sound right with this sentence, I'm not sure what the italic part means

exercises/concept/spring-cleaning/.meta/example.gleam
```diff
@@ -0,0 +1,16 @@
...
+
+pub fn remove_team_prefix(team: String) -> String {
+  let assert "Team " <> rest = team
+  rest
+}
+
+pub fn split_region_and_team(combined: String) -> #(String, String) {
+  let assert [region, "Team " <> team] = string.split(combined, ",")
+  #(region, team)
+}
```
giacomocavalieri, line 16:
```suggestion
pub fn split_region_and_team(combined: String) -> #(String, String) {
  let assert [region, team_with_prefix] = string.split(combined, ",")
  let team = remove_team_prefix(team_with_prefix)
  #(region, team)
}
```
Wouldn't it be better if we reused the previous function? Just to encourage function reuse if anyone looks into the examples as a way to learn

exercises/concept/spring-cleaning/.docs/instructions.md
lpil: Ah, I missed the "r" in "organiser"

exercises/concept/spring-cleaning/.meta/example.gleam
lpil: I think both are fine. If it was about showing the student what they can do I would prefer the current version as it shows nested patterns used with let assert, but this example isn't shown to the student so I don't think we need to think about that so much.

PR #324, opened 2023-08-15
Bit strings

lpil:


concepts/bit-strings/about.md
```diff
@@ -0,0 +1,85 @@
+# Introduction
+
+Working with binary data can be tricky, so Gleam provides a `BitString` type and accompanying syntax to construct and to pattern match on binary data.
+
+Bit string literals are defined using the `<<>>` syntax. When defining a bit string literal, it is defined in segments. Each segment has a value and and annotation, separated by a `:`. The annotation specifies how many bits will be used to encode the value, and can be omitted completely, which will default to a 8-bit integer value.
+
+```gleam
+// This defines a bit string with three segments of a single bit each
+<<0:1, 1:1, 0:1>>
+```
```
giacomocavalieri, line 10: Maybe we could add the same example leaving out the annotation like you described and explain the difference:

```suggestion
```gleam
// This defines a bit string with three segments of a single bit each
<<0:1, 1:1, 0:1>>

// This defines a bit string with three segments of 8 bit each
<<0, 1, 0>>
```
```

File: concepts/bit-strings/about.md [hunk 2]
```diff
@@ -0,0 +1,85 @@
...
+
+Specifying the type as `:1` is a shorthand for writing `:size(1)`. You need to use the longer syntax if the bit size comes from a variable.
```
giacomocavalieri, line 12: Maybe this could benefit from an example as well:
```gleam
let segment_size = 4
let bit_string = <<0:size(segment_size), 3:size(segment_size)>>
```

concepts/bit-strings/about.md
```diff
@@ -0,0 +1,85 @@
...
+If the value of the segment overflows the capacity of the segment's type, it will be truncated from the left.
+
+```gleam
+<<0b1011:3>> == <<0b0011:3>>
+// -> true
+```
+
+## Prepending and appending
+
+You can both prepend and append to an existing bit string using the bit string syntax. The `:bit_string` annotation must be used on the existing bit string if it's of variable size.
```
giacomocavalieri, line 34: What does it mean to be "variable size"? Because from the example it may not be clear since it looks like `value` is a constant composed of two segments of 3 bits each

concepts/bit-strings/about.md
```diff
@@ -0,0 +1,85 @@
...
+
+```gleam
+let value = <<0b110:3, 0b001:3>>
+let new_value = <<0b011:3, value:bit_string, 0b000:3>>
+// -> <<120, 8:size(4)>>
+```
+
+## Concatenating
+
+We can concatenate bit strings stored in variables using the syntax. The `:bit_string` annotation must be used when concatenating two bit strings of variable sizes.
```
giacomocavalieri, line 44: > ... using the syntax.

Did you mean the bit-string syntax?

concepts/bit-strings/about.md
```diff
@@ -0,0 +1,85 @@
...
+```gleam
+let first = <<0b110:3>>
+let second = <<0b001:3>>
+let concatenated = <<first:bit_string, second:bit_string>>
+// -> <<49:size(6)>>
+```
+
+## Pattern matching
+
+Pattern matching can also be done to obtain values from the syntax. You have to know the number of bits for each fragment you want to capture, with one exception: the `:bit_string` annotation can be used to pattern match on a bit string of an unknown size, but this can only be used for the last fragment.
```
giacomocavalieri, line 55: > Pattern matching can also be done to obtain values from the syntax.

Did you mean "... to obtain values from the bit string"?

exercises/concept/dna-encoding/src/dna_encoding.gleam
```diff
@@ -0,0 +1,22 @@
+pub type Nucleotide {
+  Adenine
+  Cytosine
+  Guanine
+  Thymine
+}
+
+pub fn encode_nucleotide(nucleotide: Nucleotide) -> Int {
```
giacomocavalieri, line 8: Why is it encoding the nucleotide to an `Int` and not a `BitString`?

lpil: Why would it encode to a bit string?

giacomocavalieri: I thought I'd make sense since it's an exercise on bit strings but yeah ints make more sense here :)

lpil: Good point, let's switch

PR #327, opened 2023-08-22
Iterators concept

lpil:


exercises/concept/boutique-inventory/.docs/introduction.md
```diff
@@ -0,0 +1,39 @@
...
+
+```gleam 
+iterator.unfold(#(0, 1), fn(pair) {
+  let x = pair.0 + pair.1
+  iterator.Next(element: x, accumulator: #(pair.1, x))
+})
+|> iterator.take(6)
+|> iterator.to_list
+// -> [1, 2, 3, 5, 8, 13]
+```
```
giacomocavalieri, line 39: Maybe we could emphasise that here the iterator is an infinite sequence and `take` is needed to make sure that we do not try to consume an infinite iterator

lpil: Smart!

PR #334, opened 2023-09-02
fix tuple syntax in Tisbury Treasure README

rossvz:
The README for Tisbury Treasure Hunt shows tuples missing the `#`


giacomocavalieri:
Nice catch! Thank you 💜

PR #338, opened 2023-09-06
add `_test` suffix to "bandwagoner"'s tests

giacomocavalieri:
Briker pointed out that there were some missing tests in the Bandwagoner Exercise, the tests were actually there but we forgot the "_test" suffix :)


lpil:
Ooops! Thanks

PR #348, opened 2023-09-16
Predefine do_insert instead of do_delete in erlang_extraction

michallepicki:
Because writing the insert function is the ~second~third task, I think it's more beneficial as a hint to predefine the private do_ function there so that beginners are not stuck on the different order of arguments. And then we can let the student figure this out on their own for the delete function


lpil:
Ah, this was a mistake. There are not supposed to be any private functions included in the initial file. I'll remove it, thank you

lpil (approved):

PR #350, opened 2023-09-26
Fix typo in Valentine's Day exercise

rokf:
Hello 😄 

Here's just a tiny fix for a documentation typo I've noticed in one of the exercises.

The type should be `Int` and not `int`.


giacomocavalieri:
Thank you!

Issue #276, opened 2023-05-02
Concepts exercises


lpil:
Time to make a syllabus!

F# and Elm have been excellent references for concepts and exercises as they have similar features to Gleam.

[This page](https://exercism.org/docs/building/tracks/stories) is useful for finding exercises used by other tracks for concepts.

Here's a list of all the concepts that I have identified, using the Elixir track as a base.

- [x] anonymous-functions
- [x] basics
- [x] bitstrings
- [x] bools
- [x] case-expressions
- [x] constants
- [x] custom-types
- [x] external-functions
- [x] external-types
- [x] floats
- [x] generics
- [x] guards
- [x] ints
- [x] io
- [x] iterators
- [x] labelled-arguments
- [x] labelled-fields
- [x] let-assertions
- [x] lists
- [x] maps
- [x] modules
- [x] nil
- [x] opaque-types
- [x] options
- [x] orders
- [x] pipe-operators
- [x] phantom-types
- [x] queues
- [x] recursion
- [x] regular-expressions
- [x] results
- [x] sets
- [x] strings
- [x] tail-call-recursion
- [x] tuples
- [x] type-aliases
- [x] use-expressions

bcpeinhardt:
Looks awesome :) The Elixir track seems to have a concept on processes, pid, tasks, and genserver. If the Gleam track had something similar (I guess pid, processes, tasks, actor, and maybe even supervisor), it would be immensely helpful to beam newbies like me. There's not really an existing resource that teaches BEAM with Gleam, and I think Gleam is gonna be picking up a lot of developers who've never done BEAM before (like me). Would any of that be in scope for the track?

lpil:
I think that'd be great too! I omitted them as I wanted to get the basics of the language down first, but would be good to have for sure.

PR #352, opened 2023-09-27
Use expressions

lpil:


concepts/use-expressions/about.md
```diff
@@ -0,0 +1,70 @@
+# About
+
+In Gleam it is common to write and use higher order functions, that is functions that take other functions as arguments. Sometimes when using many higher order functions at once the code can become difficult to read, with many layers of indentation.
+
+For example, here is a function that calls several functions that return `Result(Int, Nil)`, and sums the values if all three are successful.
```
giacomocavalieri, line 5:
```suggestion
For example, here is a function that calls several functions that return `Result(Int, Nil)`, and sums the values if all four are successful.
```

concepts/use-expressions/about.md
```diff
@@ -0,0 +1,70 @@
...
+pub fn main() -> Result(Int, Nil) {
+  use a <- result.try(function1())
+  use b <- result.try(function2())
+  use c <- result.try(function3())
+  use d <- result.try(function4())
+  Ok(a + b + c + d)
+}
+```
+
+A `use` expression collects all the following statements in the block into and passes it as a callback function as the final argument to the function call. The variables between the `use` keyword and the `<-` symbol are the names of the arguments that will be passed to the callback function.
```
giacomocavalieri, line 37: "Into" may be a typo?
```suggestion
A `use` expression collects all the following statements in the block and passes it as a callback function as the final argument to the function call. The variables between the `use` keyword and the `<-` symbol are the names of the arguments that will be passed to the callback function.
```

exercises/concept/expert-experiments/.docs/instructions.md
```diff
@@ -0,0 +1,64 @@
+# Instructions
+
+Daphne has been working on a system to run and record the results of her experiments. Some of the code has become a bit verbose and repetitive, so she's asked you to write some `use` expressions to help clean it up.
+
+## 1. Define the `with_retry` function
+
+Sometimes experiments can fail due to a one-off mistake, so if an experiment fails Daphne want to retry it again to see if it works the second time.
```
giacomocavalieri, line 7:
```suggestion
Sometimes experiments can fail due to a one-off mistake, so if an experiment fails Daphne wants to retry it again to see if it works the second time.
```

giacomocavalieri:
Awesome! This will close #276, right?

lpil:
Yes!!

PR #356, opened 2023-10-03
Fixed Bool mentions in instructions

rokf:
When going through the exercises I've noticed that bools were sometimes written in the wrong way in instructions. I'm attempting to fix that in this MR.


giacomocavalieri:
Thank you! 💜

PR #365, opened 2023-10-11
added the book store exercise

massivefermion:


exercises/practice/book-store/.meta/example.gleam
```diff
@@ -0,0 +1,40 @@
...
+      list.key_set(acc, book, current + 1)
+    },
+  )
+  |> list.map(fn(item) { item.1 })
+  |> list.sort(int.compare)
+  |> calculate(0.0)
+}
+
+fn calculate(counts: List(Int), acc: Float) -> Float {
+  case counts {
```
lpil, line 20: This case expression is not exhaustive and won't compile with future versions of Gleam. What happens when there are fewer than 5 elements?

If the number of elements is fixed then this should likely be a tuple and not a list, or a function that takes 5 int arguments.

config.json
```diff
@@ -1826,14 +1781,21 @@
         "slug": "nth-prime",
         "name": "Nth Prime",
         "uuid": "fda0001e-51b5-4041-a08a-2cb7dfef0f1c",
-        "practices": [
-        ],
+        "practices": [],
```
lpil, line 1784: Could you undo the unrelated changes in this file please? Thank you

lpil (commented):
Wonderful! Thank you very much. I've left 2 small notes inline.

config.json
massivefermion: Sorry about this! I always forget about my format on save!

exercises/practice/book-store/.meta/example.gleam
massivefermion: The list would always have five elements because of the initial state of the fold. But I changed it so that the clauses are exhaustive.

lpil: If you wish to assert that the pattern matches you'll need to use `let assert`. An inexhaustive `let` will be rejected by the compiler in future.

massivefermion: good to know. done!

config.json
```diff
@@ -1834,6 +1834,15 @@
...
         ],
         "difficulty": 3
+      },
+      {
+        "slug": "book-store",
+        "name": "Book Store",
+        "uuid": "f45f7c07-465a-4fec-b7ed-263d770d580c",
+        "practices": [
+        ],
+        "prerequisites": [],
```
lpil, line 1844: Could you add some practices and prerequisites? Thank you.

lpil (approved):
Thank you

PR #366, opened 2023-10-11
Update treasure_chest.gleam

inoas-nbw:
right?


lpil (changes requested):
Sorry, what's this for? This PR introduces a type error.

lpil:
You have added an extra type parameter there `password` which is not used. You can remove it without issue.

PR #368, opened 2023-10-13
Improve the secure treasure chest exercise instructions

### LINKED ISSUE #363: Secure Treasure Chest - Error Instructions
The instructions on the `Secure Treasure Chest` exercise don't say what `Error` message (`String`) is expected to be set to.

I see two options:

1. The messages are added into the descriptions.
2. The tests are updated in such a way that the error message doesn't matter.

I'd personally prefer the second option since the error message doesn't really add anything to the exercise in my opinion 🙂

rokf:
The instructions previously didn't say what the
errors should be. You found that out after running the tests.

Fixes #363


giacomocavalieri:
Thank you!

PR #369, opened 2023-10-14
Update instructions.md to reflect expected function signature

gavinmcgimpsey:
The instructions had used the singular to describe the arguments expected by `record`, rather than the plural.


lpil (approved):
Thank you

Issue #370, opened 2023-10-15
Erlang Extraction Exercise - Deletion Hint Mistake


rokf:
Hi 😄 The deletion hint (task 4) in the Erlang Extraction exercise has a mistake.

![CleanShot 2023-10-15 at 20 53 14](https://github.com/exercism/gleam/assets/2836852/15623c50-5430-4e97-bc02-215df40da3e0)

![CleanShot 2023-10-15 at 20 56 03](https://github.com/exercism/gleam/assets/2836852/0d45ee24-72f0-4c85-bfdf-b83f34782a71)

The functions have two arguments, not three.

https://www.erlang.org/doc/man/gb_trees#delete_any-2

rokf:
The hints could mention the order of the arguments as well, just as the hint for the previous task (insert / task 3). In case that this would give away too much information it could remain as it is 😄 I'm personally fine with both options.

lpil:
Thank you

PR #371, opened 2023-10-15
Fix deletion hints for Erlang Extraction

rokf:
Hello! This is a fix for #370 😃

Delete functions take two parameters, not three. I've also added a hint for the parameter order.


lpil (approved):
Thank you

PR #374, opened 2023-10-20
Remove redundant and

rokf:
I've noticed that there are (were) a couple of redundant `and` instances (duplicated). This PR removes them.


giacomocavalieri:
Thank you! 💜

PR #380, opened 2023-10-26
configlet sync for docs and metadata

BNAndras:
I ran `configlet sync`, resulting in updated docs for 27 practice exercises and metadata for 6 practice exercises.

### DIFF
--- a/exercises/practice/bowling/.docs/instructions.md
+++ b/exercises/practice/bowling/.docs/instructions.md
@@ -23,9 +23,9 @@ There are three cases for the tabulation of a frame.
 
 Here is a three frame example:
 
-| Frame 1         | Frame 2       | Frame 3                |
-| :-------------: |:-------------:| :---------------------:|
-| X (strike)      | 5/ (spare)    | 9 0 (open frame)       |
+|  Frame 1   |  Frame 2   |     Frame 3      |
+| :--------: | :--------: | :--------------: |
+| X (strike) | 5/ (spare) | 9 0 (open frame) |
 
 Frame 1 is (10 + 5 + 5) = 20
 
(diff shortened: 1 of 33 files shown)


exercises/practice/bowling/.docs/instructions.md
```diff
@@ -50,7 +50,7 @@ For a tenth frame of XXX (three strikes), the total value is 30.
 Write code to keep track of the score of a game of bowling.
 It should support two operations:
 
-- `roll(game: Game, knocked_pins: Int)` is called each time the player rolls a ball.
+- `roll(pins : int)` is called each time the player rolls a ball.
   The argument is the number of pins knocked down.
-- `score(game: Game)` is called only at the very end of the game.
-  It returns the total score for that game, wrapped in a `Result`.
+- `score() : int` is called only at the very end of the game.
+  It returns the total score for that game.
```
giacomocavalieri, line 56: It looks like this is no longer Gleam specific, could you revert this piece back to what it was?

BNAndras: Won’t it get flagged next time we do configlet sync if I revert? My understanding is track-specific information would go in an instructions-append.md which will be displayed underneath the instructions on the site.

BNAndras: Something like exercism/pyret#51 where I added a clarification the exercise in fact was implemented with the List data type despite array being in the exercise’s name.

giacomocavalieri: That's really helpful, thank you for pointing it out! I will add the additional instructions in a new PR so we can merge this

PR #384, opened 2023-10-30
Update casing for Nucleotide Count in config.json

BNAndras:
All other exercise names are in Capitalized Case so this inconsistency stands out when reading the exercise list at https://exercism.org/tracks/gleam/exercises


giacomocavalieri:
Nice catch, thank you! 💜

PR #389, opened 2023-11-25
Fixed a type in "role-playing-game"

YannickFricke:
Before the instructions used a term called `Just` for an Option - but Gleam only uses `Some` and `None`


giacomocavalieri (approved):

giacomocavalieri:
Thank you!

PR #390, opened 2023-11-25
Added instruction for role-playing-game

YannickFricke:
The last exercise does not state anything about the players health - but the tests does.


giacomocavalieri:
Thanks! Nice catch 💜

PR #391, opened 2023-11-25
Updated "newsletter" exercise

YannickFricke:
Added the hint about "simplifile" - which should be used for all the file operations.

Before it wasn't clear that the "simplifile" module is available.

Question: Should we also add a code example for how to import "simplifile"?


lpil (approved):
Fab! Thank you

PR #393, opened 2023-11-29
Use BitArray & :bits nomenclature

patrickjaberg:
Per updates in gleam 0.32 (see https://gleam.run/news/v0.32-polishing-syntax-for-stability/), this updates the language used as follows:

* `BitString` becomes `BitArray`
* "bit string" becomes "bit array"
* `:bit_string` becomes `:bits`

Also, when referring to sub-sections of bit arrays, "segment" seems to be the term of choice, so a few occurrences of "fragment" were changes to "segment".  FWIW, "segment" is also the term used in [The Gleam Book](https://gleam.run/book/tour/bit-strings.html).


giacomocavalieri:
Thank you!

PR #397, opened 2023-12-08
Fix wording in introduction.md

patrickjaberg:
The change is probably best described by the diff.  :-)

### DIFF
--- a/exercises/concept/log-parser/.docs/introduction.md
+++ b/exercises/concept/log-parser/.docs/introduction.md
@@ -18,7 +18,7 @@ The most common way to create regular expressions is using the `regex.from_strin
 let assert Ok(re) = regex.from_string("test")
 ```
 
-The regular expression creation function returns an error if the regular expression syntax is invalid, so a let-assertion has been used here to ensure the regular expression is valid.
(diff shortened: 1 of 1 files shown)


lpil (commented):
This was right before I'm afraid. There are 2 functions that can be used to create Regex objects

evanasse:
Should it be "The regular expression creation functions return an error...", then?
The word "return" is missing in the sentence.

exercises/concept/log-parser/.docs/introduction.md
```diff
@@ -18,7 +18,7 @@ The most common way to create regular expressions is using the `regex.from_strin
 let assert Ok(re) = regex.from_string("test")
 ```
 
-The regular expression creation functions an error if the regular expression syntax is invalid, so a let-assertion has been used here to ensure the regular expression is valid.
+The regular expression creation function returns an error if the regular expression syntax is invalid, so a let-assertion has been used here to ensure the regular expression is valid.
```
lpil, line 21:
```suggestion
The regular expression creation functions return an error if the regular expression syntax is invalid, so a let-assertion has been used here to ensure the regular expression is valid.
```

lpil (approved):
Thank you both!

PR #398, opened 2023-12-20
Update instructions.md

sotolf2:
Updated the instructions to better fit the task given in the tests, since the task as written in the readme was not really fitting to the task actually given in the src/test of the task


lpil (commented):
Hello! This change would need to be made here: https://github.com/exercism/problem-specifications/blob/main/exercises/zipper/description.md

PR #403, opened 2024-01-16
leap: sync

ErikSchierboom:
Sync the `leap` exercise with the latest data, as defined in https://github.com/exercism/problem-specifications/tree/main/exercises/leap.


lpil (approved):
Thank you!

PR #409, opened 2024-01-28
Sieve

bcpeinhardt:
I hope I did this right! Let me know if I'm missing a file or something.


bcpeinhardt:
Hey look at that, the CI passed, fabulous :)

exercises/practice/sieve/src/sieve.gleam
```diff
@@ -0,0 +1,3 @@
+pub fn primes_up_to(upper_bound upper_bound: Int) -> List(Int) {
```
lpil, line 1: Could you remove this label please

exercises/practice/sieve/.meta/example.gleam
```diff
@@ -0,0 +1,50 @@
...
+  let possible_primes = list.range(from: 2, to: upper_bound)
+
+  // Run the sieve algorithm on the list of possible primes.
+  sieve([], [], possible_primes, upper_bound)
+  |> list.reverse
+}
+
+fn sieve(
+  primes: List(Int),
+  composites: List(Int),
```
lpil, line 19: I think this wants to be a set rather than a list. Appending to and finding elements in a list are expensive operations.

exercises/practice/sieve/.meta/example.gleam
```diff
@@ -0,0 +1,50 @@
...
+  candidates: List(Int),
+  upper_bound: Int,
+) -> List(Int) {
+  // BASE CASE: If there are no more candidates, we're done.
+  use <- bool.guard(when: list.is_empty(candidates), return: primes)
+  let assert [next, ..rest] = candidates
```
lpil, line 25: The assertion here can be avoided using pattern matching. Whenever possible code should be written in a style that uses static analysis to ensure data is handled correctly rather than doing procedural two-step check-and-then-assert patterns, as the compiler will let you know if there's a mistake.

lpil (commented):
Thank you! I've left some small notes

bcpeinhardt:
How's that?

lpil (approved):
Beautiful! Thank you

PR #424, opened 2024-03-08
Update instructions.md

badurinantun:
The instructions were ambiguous and led to an expectation of receiving a list of tuples as the output.


lpil (approved):
Thank you!

Issue #396, opened 2023-12-05
Update `map` exercises to `dict`


mcastorina:
Some of the exercises teach about the [map](https://hexdocs.pm/gleam_stdlib/gleam/map.html) data structure, which is deprecated in `v0.33.0` in favor of [dict](https://hexdocs.pm/gleam_stdlib/gleam/dict.html).

Should these exercises be updated?

PR #426, opened 2024-03-12
Add parallel letter frequency

kahgoh:
Adds the [parallel letter frequency for 48in24](https://forum.exercism.org/t/48in24-exercise-03-12-parallel-letter-frequency/9927/).


config.json
```diff
@@ -1827,6 +1827,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 4
+      },
+      {
+        "slug": "parallel-letter-frequency",
+        "name": "Parallel Letter Frequency",
+        "uuid": "a9e52150-65bc-44a9-9ef6-95b995aea794",
+        "practices": [],
+        "prerequisites": [],
```
lpil, line 1836: We'll need to have some practices and prerequisites here!

lpil (approved):
Wonderful! One small note inline

exercises/practice/parallel-letter-frequency/.meta/example.gleam
```diff
@@ -0,0 +1,49 @@
+import gleam/dict.{type Dict}
+import gleam/list
+import gleam/option.{type Option, None, Some}
+import gleam/otp/task
```
lpil, line 4: This hasn't been taught yet by this point, and it's an indirect dependency so it'll emit a warning. Before OTP can be used lessons will need to be added to teach it.

config.json
```diff
@@ -1827,6 +1827,19 @@
...
+      {
+        "slug": "parallel-letter-frequency",
+        "name": "Parallel Letter Frequency",
+        "uuid": "a9e52150-65bc-44a9-9ef6-95b995aea794",
+        "practices": [
+          "strings",
+          "lists"
+        ],
+        "prerequisites": [
+          "lists"
```
lpil, line 1840: This is missing regex, options, strings, dicts, and OTP tasks.

kahgoh: Ah I think missed those earlier as I didn't seem them listed in the existing concepts. I've just added them and pushed.

With the `dict` one, I put `maps` in its place as I noticed the existing ones called it `maps` instead of `dict` and I assumed it would be updated as part #396. Let me know if you'd prefer me to change it to `dict` instead.

kahgoh: Update - Nevermind, I just saw pull request #431 and decided to go ahead and this pull request to use `dicts` to match.

exercises/practice/parallel-letter-frequency/.meta/example.gleam
kahgoh: I noticed the CI build failed as I had added the OTP task to the prerequisite, although it hasn't been added yet. I've since it removed it and raised #454.

PR #430, opened 2024-03-13
Update introduction.md - Provide "type" in Import

daniellionel01:
Not having the "type" prefixing the import of "Option" gives me an error in v1.0


giacomocavalieri:
Thank you!!

PR #431, opened 2024-03-14
Update Gleam concepts: replace maps with dicts

Guria:
Follow up missed bits for faaed22


Guria:
Note sure how concept rename could affect the course

lpil (commented):
Brilliant! Thank you!

We'll need to also update the `config.json` file which declares and references the maps concept

Guria:
updated

lpil (approved):
Thank you!!

PR #434, opened 2024-03-21
Add missing test for low-level full mana requirement `Role Playing Game`

lennnerd:
The problem description states: `If they player's level is below 10, their mana should be None`. One of the top results did not implement this requirement due to a missing test. 

This PR modifies a test verifying that low level players get their `mana` set to `None`.

@Reviewer: If preferred, i can also add a new test instead modifying the current test.


lpil:
Perhaps removing that requirement would be best. It's less disruptive and it doesn't teach anything in particular.

lennnerd:
Fair point.

Personally, as a beginner I quite liked the aspect of trying to reduce nesting with the extra mana requirement. But the less disruptive change is changing the docs. Its your call. I adapted the PR.

lpil (approved):
Thank you! 💜

PR #436, opened 2024-03-25
Update instructions.md

evanasse:
Replace camel case with snake case for variable name `new_coach`.

I was also wondering if we should change the function call `replace_coach team new_coach` to match the other function calls, i.e., change to `replace_coach(team, new_coach)`.

I didn't make the change because the current syntax is valid Gleam. 
Confirmed with this solution of the exercise: https://exercism.org/tracks/gleam/exercises/bandwagoner/solutions/evanasse)


evanasse:
Of course, there you go!

lpil (approved):
Thank you!

PR #438, opened 2024-03-27
Corrected argument typo

alex-sarson:
If the intended return for the function is 10 then the argument in the divider function should be 30.


lpil (commented):
Isn't this correct as it's int division rather than float division?

alex-sarson:
You're absolutely right, I was thinking in a float division sense without realising. I'll close this

PR #440, opened 2024-03-28
Update instructions.md for high-school-sweetheart exercise

evanasse:
Change function notation from `pair/2` to `pair`.

I haven't seen this function notation for Gleam in the Exercism track so far nor in the [Gleam language tour](https://tour.gleam.run/) and I only know about Gleam since v1.0.
So I don't know if it's a pre-v1.0 exercise artefact and the notation got removed or if it's something that should be presented prior to this exercise to avoid confusion about the "/2".


giacomocavalieri:
Thank you! Yeah that's an Erlang thing

PR #441, opened 2024-03-28
Update grade_school.gleam src to match the description and the tests

LiorKovalio:


lpil (commented):
Fab! Thank you. Could you run the formatter please.

lpil (approved):
Thank you

PR #444, opened 2024-03-30
fix role_playing_game test according to instructions

GioeleBucci:
The exercise's instructions say: "If they player's level is below 10, their mana should be untouched."

Since the test checked against `None` there were unintended solutions that did not account for the player level, such as:

`Some(Player(..player, health: 100, mana: option.map(player.mana, fn(_){ 100 })))`

### DIFF
--- a/exercises/concept/role-playing-game/test/role_playing_game_test.gleam
+++ b/exercises/concept/role-playing-game/test/role_playing_game_test.gleam
@@ -26,9 +26,11 @@ pub fn revive_a_player_that_is_alive_should_return_none_test() {
 }
 
 pub fn reviving_a_low_level_player_resets_its_health_to_100_test() {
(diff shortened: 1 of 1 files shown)


lpil (commented):
Hello! Isn't this unchanged and thus correct both before and after this change?

GioeleBucci:
> Hello! Isn't this unchanged and thus correct both before and after this change?

It is indeed correct, but the test uses `None`, which is an unfortunate value to check against since it allows defaulting to `None`, independently from the previous value of the "mana".
If the value to keep was, for example 7, one could ignore it and set it to `None`: since the test checks against that same value it would pass when it shouldn't. Changing the value the test uses to an integer would prevent such behaviour.
I don't know if I explained it in the best terms but hopefully it's understandable.

PR #449, opened 2024-04-03
changed: simplify a phrase to improve clarity

dalprahcd:
Consider simplifying the phrase to improve clarity. For example, "add an additional" can be redundant. But in the end both phrases are fine, you can choose which one to keep it.


lpil (approved):
Thank you

PR #451, opened 2024-04-03
changed to use dash for markdown unordered list

dalprahcd:
I noticed some places where using both `*` and `-` to represent unorderd list, I believe the default being used is `-`


lpil (approved):
Thank you

PR #472, opened 2024-05-02
align function naming in instructions for betty's bike shop

nils-geistmann:
The instructions for `betty's bike shop` refer to a function named `pounds_to_pence` which doesn't exist in this exercise. To correct name is `pounds_to_string`.


lpil (approved):
Thank you!

PR #475, opened 2024-05-07
Add zebra-puzzle exercise

glennj:
I'm sure there's an "eleganter" way solve this, but ...


config.json
```diff
@@ -1833,6 +1833,14 @@
           "recursion"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "zebra-puzzle",
+        "name": "Zebra Puzzle",
+        "uuid": "82a312b6-5d71-4119-b3b0-211d572c17c8",
+        "practices": [],
+        "prerequisites": [],
```
lpil, line 1842: Could you add the practices and prerequisites please

lpil (commented):
Thank you!

Looking at the other languages this would normally be a function for each of the water drinker and the zebra owner. How come you have opted for a data structure holding both?

config.json
glennj: Will do.

glennj:
> Looking at the other languages this would normally be a function for each of the water drinker and the zebra owner. How come you have opted for a data structure holding both?

Once the puzzle is solved (once), both answers are known. Having to solve it twice is redundant.

lpil:
Please switch over to having two functions, thank you

PR #476, opened 2024-05-09
Fix total_yen_test case in sticker-shop exercise

airclovis:
This PR is updating the `total_yen_test` case to use Yen currency.

For now the test case is using Euro currency while we should check if the `sticker_shop.total` function is working as expected with Yen currency.


lpil (approved):
Thank you!!

PR #478, opened 2024-05-11
Rename Bit Strings to Bit Arrays

PastMoments:
Gleam updated to rename bit strings to bit arrays, but the exercism pages hadn't updated yet.

This PR just renames all mentions of bit strings to bit arrays. It also renames `:bit_string` syntax to `:bits`


lpil (approved):
Thank you

PR #489, opened 2024-06-07
Replaced the word 'Maybe' with 'perhaps to fix error "error: Shell command failure"

KimBrusevold:
As discussed in Discord (https://discord.com/channels/768594524158427167/1247980115091128423/1247984729051893780) (https://discord.com/channels/768594524158427167/1248553138428969010) there is an error with the valentines-day exercism exercise

We received the following error when we ran your code:

error: Shell command failure
There was a problem when running the shell command escript.

Made a quick change to the exercise to replace the word "Maybe" with "Perhaps". Maby a hacky way of fixing it, but don't have to deal with build and/or config files. 

I don't know if this breaks the exercise for those that already has completed it. Hopefully someone else can say something about this

### DIFF
--- a/exercises/concept/valentines-day/.docs/instructions.md
+++ b/exercises/concept/valentines-day/.docs/instructions.md
@@ -14,9 +14,9 @@ You have six tasks to help choose your Valentine's day activity.
 
 ## 1. Define the approval
 
-For each idea your partner proposes, you respond with one of three options: yes, no or maybe.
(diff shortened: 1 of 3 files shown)


KimBrusevold:
I guess maybe this shouldn't be merged, as this would invalidate all completed tests. Perhaps fiddling with some build scripts is the only way.

PR #494, opened 2024-06-12
Changed Maybe to Possibly in Approval

FJSolutions:
Changed the 'Maybe' option of the 'Approval' type to 'Possibly' to deal with issue #3233


lpil:
Hello! We have rolled back to OTP26 so this is no longer an issue.

PR #499, opened 2024-07-02
Fix Monday bonus hint

Varpie:
The hint for the last part of the High Score Board exercise suggests using `dict_values` from the gleam/dict library.

The `dict_values` function doesn't exist on gleam/dict, but `map_values` exists, and looking at the description and community solutions, it seems to be what was intended in the first place.


giacomocavalieri:
Oh my, this must have been a find and replace gone wrong from when we changed the `map` module to `dict` 😁

Issue #492, opened 2024-06-10
reference to fsharp in instructions markdown


daneroe:
Very minor, but noticed while doing the exercise locally

https://github.com/exercism/gleam/blob/b321e3afc51c3bb41ca6a469ffeae72e5c618c5d/exercises/concept/bandwagoner/.docs/instructions.md?plain=1#L82C1-L84C1

by extension, colours for that code block are rendering a bit off on the website

https://exercism.org/tracks/gleam/exercises/bandwagoner

Great site and track (and language) BTW!

Varpie:
Corrected in #500, which has been merged. I don't think that I can close this issue, but you can verify it and close it yourself.

daneroe:
Awesome, thanks :)

PR #500, opened 2024-07-02
Updated Markdown code block language

Varpie:
As noted in  #492, a Markdown code block uses the F# language instead of Gleam, which results in an inconsistent styling.


giacomocavalieri:
Nice catch!! Thank you so much

PR #506, opened 2024-07-08
Fix code example and add missing link

Varpie:
The code example for the `add_card` function instructions uses invalid syntax.
This change fixes it to use the correct parameters.

In addition, there is a broken link in one of the hints.


giacomocavalieri:
Thank you!

PR #512, opened 2024-07-12
Update instructions.md

probablyjg:
The instructions don't quite follow the same format as other exercises: `file_name.function_name` vs `function_name`.

Passing the tests of exercise 3 requires that an empty list returns false, which is non-obvious (no birds record ~ [0]?)

Maybe change the test, or change the docs to reflect this


lpil (approved):
Thank you

PR #514, opened 2024-07-16
docs(erlang-extraction): Add hints that point toward docs

coronary:
I felt that the hints on 'Erlang Extraction' weren't direct enough in hinting that the external functions had a different order than the predefined functions used for testing. Despite the current hints that allude to a different order I still didn't quite get the picture so I wanted to add something that was a little more direct.


lpil (approved):
Nice, thank you!

PR #517, opened 2024-07-19
update test case for rna transcription

zhuangya:


lpil (commented):
Hello! What's the motivation for this change?

zhuangya:
Hi Louis @lpil, the reason I am raising this PR is that the name of this test case does not match its content, and it is also identical to the [next one](https://github.com/exercism/gleam/blob/5d83a6a061760df465a9c2aa7a96943e2aa9c684/exercises/practice/rna-transcription/test/rna_transcription_test.gleam#L15-L25).

lpil:
Ah thank you! Does it now match what is specified in https://github.com/exercism/problem-specifications ?

zhuangya:
@lpil yeah, totally. let me put a [spec ref](https://github.com/exercism/problem-specifications/blob/685ec55d9388937bbb3cc836b52b3ce27f208f37/exercises/rna-transcription/instructions.md) here for convenience

lpil (approved):
Thank you!

PR #518, opened 2024-07-30
Adding documentation links

Varpie:
Most exercises include a link to the standard library documentation when they suggest using a specific function.  
The Erlang Extraction exercise was missing those links.


lpil (approved):
Thank you

PR #522, opened 2024-08-22
Updated gotta-snatch-em-all intro docs

ddominguez:
This update removes an example set.delete() with an incorrect return value.

I thought it would be better to remove the example instead of fixing the return value since the `set.contains` and `set.size` examples are expecting `set.from_list([88])`.


lpil (approved):
Thanks

PR #534, opened 2024-11-28
(lasagna) Remove ambiguity in example for `total_time_in_minutes`

M1N0RM1N3R:
According to the example before,
```gleam
pub fn total_time_in_minutes(layers: Int, elapsed: Int) {preparation_time_in_minutes(layers) + remaining_minutes_in_oven(elapsed)}
```
and
```gleam
pub fn total_time_in_minutes(layers: Int, elapsed: Int) {preparation_time_in_minutes(layers) + elapsed}
```
would both be valid, but only the latter is correct.


lpil (approved):
Thanks

PR #541, opened 2025-02-04
Update wording explaining `Gt` and `Lt`

sisou:
Explain the abbreviations `Gt` and `Lt` with the actual words they abbreviate ("greater than" and "lower than" respectively).


lpil (approved):
Thank you

Issue #542, opened 2025-02-08
Concepts use removed modules from stdlib


fridewald:
The current version (0.54.0) of the gleam stdlib has removed some modules that are referenced in some exercises and the learn-track

* gleam/regex is moved to gleam/regexp in the https://hexdocs.pm/gleam_regexp/ package
* gleam/queue is moved to gleam/deque in the https://hexdocs.pm/gleam_deque/ package
* gleam/iterator is moved to gleam/yielder in the https://hexdocs.pm/gleam_yielder/ package

This can lead to confusion if somebody just learning gleam will learn these concepts but then when using the current version of the stdlib can not find these modules anymore.

We could decide for each concept if we want to update the exercises to use the current version of stdlib and the new modules or removing the concept as it is not part of the stdlib?

Affected exercises & concepts:
* regex:
  * https://github.com/exercism/gleam/tree/main/concepts/regular-expressions
  * https://github.com/exercism/gleam/blob/main/exercises/practice/simple-cipher/test/simple_cipher_test.gleam
* queue:
  * https://github.com/exercism/gleam/tree/main/concepts/queues
  * https://github.com/exercism/gleam/tree/main/exercises/concept/magician-in-training
* iterator: 
  * https://github.com/exercism/gleam/tree/main/concepts/iterators
  * https://github.com/exercism/gleam/tree/main/exercises/concept/boutique-inventory
  * https://github.com/exercism/gleam/blob/main/exercises/practice/simple-cipher/test/simple_cipher_test.gleam
  * https://github.com/exercism/gleam/blob/main/exercises/practice/dnd-character/test/dnd_character_test.gleam

fridewald:
In #537 @lpil suggest to remove the iterator lesson.

lpil:
Removing them would be great!

PR #543, opened 2025-02-08
Update removed stdlib modules

fridewald:
#542 Remove concepts that focus on removed modules from the stdlib.

I moved the associated exercises to the practice folder as exercises cannot be removed. https://exercism.org/docs/building/tracks/deprecated-exercises


config.json
```diff
@@ -485,6 +449,38 @@
...
+      {
+        "slug": "magician-in-training",
+        "name": "Magician in Training",
+        "uuid": "d7a5dc48-3a0f-4953-a450-67f5f169efba",
+        "practices": [
+        ],
+        "prerequisites": [
+          "opaque-types"
+        ],
+        "status": "deprecated"
```
fridewald, line 472: I'm not sure if the exercises should be deprecated or not.

lpil: Deprecation sounds good to me.

lpil (commented):
Thank you!

This will need to be merged at the same time as the packages versions and Gleam versions are updated, so we'll need to update all the exercises rather than just these ones.

fridewald:
@lpil So just to understand this correctly, the approach would be 
* to update the `exercism_test_runner` to the latest stdlib version
* release the package with a new version
* merge this pr (with all exercises updated to latest stdlib)

For me, that would mean I can prepare the updates in the `gleam-test-runner` repo and here? And you'd have to do the synchronized merge?

fridewald:
I updated the exercises and examples to use the newer stdlib. To test I locally run the `bin/test` script using the changes from the other PR. 

If I understand it correctly, after the [runner PR](https://github.com/exercism/gleam-test-runner/pull/83) is merged, we can run the `bin/sync-exercise-project-configs` script to update the all the exercise packages?

Hope this will then run smoothly as the examples are already updated

lpil:
Yes that's right! Sorry for the delay.

fridewald:
I updated the project configs for the exercises based on the files in https://github.com/exercism/gleam-test-runner/pull/83

lpil (approved):
Thank you!!!

PR #544, opened 2025-02-20
docs(concepts): update sets example to show correct outcome

Joxtacy:
I just found a slight inconsistency in the docs for the `Sets` concept.


lpil (commented):
This was correct before! It's now saying that if you have a set containing `88` and delete `88` then it'll somehow contain `89`, but that wasn't in the set at all.

Joxtacy:
> This was correct before! It's now saying that if you have a set containing `88` and delete `88` then it'll somehow contain `89`, but that wasn't in the set at all.

You're right. I was not reading it properly. 🙃

PR #548, opened 2025-03-13
Sort exercises

keiravillekode:
Exercises are sorted by numeric difficulty, then slug.

https://forum.exercism.org/t/gleam-track-overview-not-sorted-by-difficulty/16331


keiravillekode:
If the track maintainers prefer the current order, I understand.

PR #550, opened 2025-04-27
fix gleam configuration files

fridewald:
Fix gleam configuration files based on git tags instead of main branch.

Related to the release of #542 and https://github.com/exercism/gleam-test-runner/pull/83.

When one updates the gleam test runner it can cause the exercise solutions to not compile anymore. With this, the process would be independent. The downside is that one would also need to update the scripts when releasing an new rest runner version.

### DIFF
--- a/bin/sync-exercise-project-configs
+++ b/bin/sync-exercise-project-configs
@@ -31,7 +31,7 @@ check() {
   project="$1"
   update_config_name "$project"
 
-  if ! diff "$canonical_config" "$project"/gleam.toml || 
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you

lpil (changes requested):
Sorry, I misread this. 

This current behaviour is the intended behaviour. Supporting other versions of the runner is not supported and not something we aim to encourage.

fridewald:
Oki, then I close this PR

PR #569, opened 2025-09-14
UPDATE: Change to correct float comparisons

clouby:
I noticed that we were using using the wrong expression for comparisons on Float types on Gleam, replace `>` to `>.`


giacomocavalieri:
Thank you!

PR #581, opened 2026-02-15
Tilsbury Treasure Hunt doesn't test for the first case of section 4

zeddash:
> The Brass Spyglass can be swapped for any other treasure at the Abandoned Lighthouse.

This was completely untested for and could be skipped.


khalidbelk:
Thanks for finding out ! @zeddash. Indeed, this explicitly makes the user handle that case which can currently be omitted.

khalidbelk (approved):

khalidbelk:
It's been tested on my side. Can be merged ;) @lpil (srry for all the pings btw 😭, but off the top of my head I don't see any other maintainer with the rights here)

khalidbelk:
@lpil ; )

IsaacG:
+cc @exercism/gleam

IsaacG:
+cc @exercism/gleam

Issue #585, opened 2026-05-14
Out of date info in Bit Array concept


RodideBoer:
In `concepts/bit-arrays/about.md` the last part talks about how Bit Arrays may be displayed as strings with functions such as `io.debug`, which is no longer part of Gleam. I'm not sure in what way Bit Arrays will be printed as strings; I tried `echo`, but that just displays as a Bit Array, and `io.print` / `io.println` do not accept BitArray.

So it might just be removed, or it should be changed into what does still work.

The part I'm talking about:

> If you create a bit array that represents a printable UTF-8 encoded string, it may displayed as a string by functions such as io.debug. This is due to an implementation detail of how Gleam represents strings internally.

```
<<>>
// -> ""

<<65, 66, 67>>
// -> "ABC"
```

lpil:
Thank you.

Exercism runs Gleam on Erlang, and it sounds like you're using JavaScript. They use different runtime representations which means that `echo` will never print bit arrays as strings.

RodideBoer:
Oh right. I was quickly testing some stuff in the browser on tour.gleam.run in stead of doing it properly in a real project, so that makes sense. I just tested it and `echo` does work on Erlang.

I am still correct about that last part being outdated right? `io.debug` isn't available anymore if I'm not mistaken. And `io.print` and `io.println` do not accept `BitArray`s.

lpil:
Yeah for sure. We could def improve the text here. Swapping `io.debug` for `echo` and making note of that behaviour being on Erlang would be good

RodideBoer:
I created a pull request for this: https://github.com/exercism/gleam/pull/586

PR #586, opened 2026-05-22
Update info on BitArray (issue #585)

RodideBoer:


concepts/bit-arrays/about.md
```diff
@@ -82,7 +82,7 @@ By default, bit arrays are displayed in fragments of 8 bits (a byte), even if yo
 // -> <<251, 3:size(3)>>
 ```
 
-If you create a bit array that represents a printable UTF-8 encoded string, it may displayed as a string by functions such as `io.debug`. This is due to an implementation detail of how Gleam represents strings internally.
+If you create a bit array that represents a printable UTF-8 encoded string, it may displayed as a string by using the `echo keyword` when targeting Erlang. This is due to an implementation detail of how Gleam represents strings internally. Note: this does not work when targeting Javascript.
```
lpil: Given the "when targetting Erlang" I think we could drop the bit about JavaScript? "does not work" makes it sound like it's broken

RodideBoer: I think you're right. I added it at the last moment to be complete, but I don't think it needs it and it indeed might sound like it's broken.

RodideBoer: Done.

lpil (approved):
Thank you!!

